PR #36, opened 2015-02-03
Updating .kproj files

BrennanConroy:
@muratg


samples/MemoryCacheConcurencySample/MemoryCacheConcurencySample.kproj
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
```
Praburaj: Will be good to change it to `.xproj` since you are anyway touch this.

PR #51, opened 2015-03-13
Using [NotNull] from Common repo package

Praburaj:
@muratg


src/Microsoft.Framework.Caching.Distributed/CacheExtensions.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.IO;
+using Microsoft.Framework.Internal;
```
muratg: Is this required here?  Nothing else seems to have changed in this file.

src/Microsoft.Framework.Caching.Distributed/LocalCache.cs
```diff
@@ -4,6 +4,7 @@
 using System;
 using System.IO;
 using Microsoft.Framework.Caching.Memory;
+using Microsoft.Framework.Internal;
```
muratg: Ditto

muratg:
:shipit:

src/Microsoft.Framework.Caching.Distributed/CacheExtensions.cs
Praburaj: Yes the [NotNullAttribute] in the common package lives in this namespace. That's why I added this namespace after removing the `NotNullAttribute.cs` file in each of the projects.

muratg: :+1:

PR #53, opened 2015-03-31
Target netcore451

bricelam:
This is part of aspnet/EntityFramework#1952


davidfowl:
specifically netcore451? windows store 8.1? Why?

bricelam:
Unfortunately, that's what the preview tooling uses under the covers. We'll update as they update. Ultimately we'll probably just use core50 when everything supports it.

ajcvickers:
:shipit:

PR #64, opened 2015-05-05
Using 'nameof' operator instead of magic strings

hishamco:
Using the "nameof" operator instead of magic string will help us if the variable change in the future for whatever the reason


hishamco:
@Eilon Can you merge this fix like the other repos?

hishamco:
Thanks

PR #70, opened 2015-06-11
[Design] Implementation of SQL Server Cache

kichalla:
@Tratcher @Eilon @muratg  @lodejard

Please note that I do not have tests yet, but I have verified that the following is working end-to-end with different samples(including the one in Session repo : https://github.com/aspnet/Session/pull/34)


src/Microsoft.Framework.Caching.SqlServer/SqlServerCache.cs
```diff
@@ -0,0 +1,315 @@
...
+                    //TODO: GetFieldValueAsync?
+                    oldExpirationTimeUTC = DateTimeOffset.Parse(sqlDataReader[ExpiresAtTimeUTC].ToString());
+                    sessionvalue = (byte[])sqlDataReader[SessionValue];
+                }
+                else
+                {
+                    return null;
+                }
+
+                // Check if the session is going to expire within the next timeout period(ex:20 minutes),
```
Tratcher: Remove comments about session in the cache component.

src/Microsoft.Framework.Caching.SqlServer/SqlServerCache.cs
```diff
@@ -0,0 +1,315 @@
...
+
+                await connection.OpenAsync();
+
+                var effectedRowCount = await command.ExecuteNonQueryAsync();
+            }
+
+            StartScanForExpiredSessions();
+        }
+
+        public void Set([NotNull] string key, [NotNull] byte[] value, DistributedCacheEntryOptions options)
```
Tratcher: Why do you ignore DistributedCacheEntryOptions AbsoluteTimeout and SlidingExpiration? IdleTimeout should not be a global setting for the cache, it should be set by the caller on a per key basis.  Compare with the Redis cache.

src/Microsoft.Framework.Caching.SqlServer/Program.cs
```diff
@@ -0,0 +1,78 @@
...
+
+using System;
+using System.Data;
+using System.Data.SqlClient;
+
+namespace Microsoft.Framework.Caching.SqlServer
+{
+    public class Program
+    {
+        private const string DefaultDatabaseName = "ASPNET5SessionState";
```
Tratcher: Remove all references to Asp.Net and Session in the Sql cache.

kichalla: :+1:

davidfowl:
Why are there so many references to session in the cache implementation? Isn't session decoupled from the distributed cache?

MichaCo:
@davidfowl agree, may I also ask you if you think it really makes sense to have a database based distributed cache? Wouldn't we not simply use EF or something else to store something in the database instead of treating it as cache layer?

src/Microsoft.Framework.Caching.SqlServer/SqlServerCache.cs
kichalla: :+1:

src/Microsoft.Framework.Caching.SqlServer/SqlServerCache.cs
kichalla: yeah :+1:  ...(probably this is something related to my earlier discussion about creating just a 'store' for session and not get tied up with caching...this does not feel intuitive intuitive to me actually...like saying `AddSqlServerCache` and then `AddSession`)

kichalla:
@davidfowl , @Tratcher : I have updated the PR now. Please take a look.

samples/SqlServerCacheConcurencySample/Program.cs
```diff
@@ -0,0 +1,112 @@
...
+        public void Main()
+        {
+            _cacheEntryOptions = new DistributedCacheEntryOptions();
+            _cacheEntryOptions.SetSlidingExpiration(TimeSpan.FromSeconds(10));
+
+            var loggerFactory = new LoggerFactory();
+            loggerFactory.AddConsole();
+            var cache = new SqlServerCache(new SqlServerCacheOptions()
+            {
+                ConnectionString = "Server=localhost;Database=ASPNET5CacheTest;Trusted_Connection=True;",
```
Tratcher: I reccomend removing references to AspNet5. This is a Ms.Framework component, not an AspNet component.

samples/SqlServerCacheConcurencySample/project.json
```diff
@@ -0,0 +1,19 @@
+{
+    "commands": {
+        "install-sqlservercache": "Microsoft.Framework.Caching.SqlServer",
+        "SqlServerCacheConcurencySample": "SqlServerCacheConcurencySample"
+    },
+    "dependencies": {
+        "Microsoft.Framework.Caching.SqlServer": "1.0.0-*",
+        "Microsoft.Framework.Logging": "1.0.0-*",
```
Tratcher: Transitive dependency

samples/SqlServerCacheSample/project.json
```diff
@@ -0,0 +1,19 @@
+{
+    "commands": {
+        "install-sqlservercache": "Microsoft.Framework.Caching.SqlServer",
```
Tratcher: `install`? What does this operation actually do? create tables?

Also, what happens if you run this twice? I assume it will fail?

Tratcher:
You should find someone from the EF team to look at your SQL usage.

src/Microsoft.Framework.Caching.SqlServer/SqlServerCache.cs
```diff
@@ -0,0 +1,395 @@
...
+        {
+            _options = options.Options;
+            _systemClock = _options.SystemClock;
+            _sqlQueries = new SqlQueries(_options.TableName);
+            _logger = loggerFactory.CreateLogger<SqlServerCache>();
+        }
+
+        public void Connect()
+        {
+            //TODO:
```
Tratcher: How about performing a no-op operation to verify the connection string and server availability?

kichalla: yeah, i thought about it but was wondering if we should have this method at all...like we discussed earlier, we would be sequencing the deployment of the database, then the web server etc...if the database was not up and ready before the web app starts, the web app would fail..

samples/SqlServerCacheSample/project.json
kichalla: yes, this creates tables as of now...i didn't do much polish here as there is other idea that came up yesterday about creating tables at run time...

samples/SqlServerCacheConcurencySample/Program.cs
kichalla: Sure

samples/SqlServerCacheConcurencySample/project.json
kichalla: Are you suggesting that reference to `Console` logger alone should be enough?...I need reference to `LoggerFactory` too (to which I add the ConsoleLoggerProvider) which is part of `Logging` package.

src/Microsoft.Framework.Caching.SqlServer/SqlServerCache.cs
Tratcher: Hmm, open a separate [needs design] bug for Connect then.

samples/SqlServerCacheConcurencySample/project.json
Tratcher: Oh, nevermind. Console depends on Abstractions, not Logging directly.

src/Microsoft.Framework.Caching.SqlServer/SqlServerCache.cs
```diff
@@ -0,0 +1,395 @@
...
+using Microsoft.Framework.Caching.Distributed;
+using Microsoft.Framework.Internal;
+using Microsoft.Framework.Logging;
+using Microsoft.Framework.OptionsModel;
+
+namespace Microsoft.Framework.Caching.SqlServer
+{
+    public class SqlServerCache : IDistributedCache
+    {
+        internal const int ExpirationTimeMultiplier = 2;
```
Tratcher: Needs a comment on why the expiration time is being multiplied.

src/Microsoft.Framework.Caching.SqlServer/SqlServerCache.cs
kichalla: Not sure if this is a bug with `Connect` itself as its the usage in SessionMiddleware that can cause the issue...i will open a bug in Session repo?

src/Microsoft.Framework.Caching.SqlServer/SqlServerCache.cs
```diff
@@ -0,0 +1,395 @@
...
+                {
+                    if (utcNow >= (oldExpirationTimeUTC - slidingExpiration.Value))
+                    {
+                        reader.Dispose();
+
+                        command = new SqlCommand(_sqlQueries.UpdateCacheItemExpiration, connection);
+                        command.Parameters.AddWithValue(CacheItemId, key);
+                        command.Parameters.AddWithValue(
+                            ExpiresAtTimeUTC,
+                            GetNewExpirationTime(slidingExpiration.Value, utcNow));
```
Tratcher: If AbsoluteExpiration is set you need to account for that here. SlidingExpiration should not renew beyond AbsoluteExpiration.

src/Microsoft.Framework.Caching.SqlServer/SqlServerCache.cs
Tratcher: Sure

src/Microsoft.Framework.Caching.SqlServer/SqlServerCache.cs
kichalla: Ah ok...thought user should use only one of them(now i remember seeing some code in MemoryCache related to this).

src/Microsoft.Framework.Caching.SqlServer/SqlServerCache.cs
```diff
@@ -0,0 +1,395 @@
...
+            try
+            {
+                connection.Open();
+
+                var effectedRowCount = command.ExecuteNonQuery();
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError("Error deleting expired cache items", ex);
+                //TODO: should we throw exception here...if yes, will it be caught as this is a background thread
```
Tratcher: Don't throw from background tasks. Logging is your best option.

kichalla:
Updated. /cc: @Tratcher, @divega 

@divega : I will discuss with you offline regarding this PR.

src/Microsoft.Framework.Caching.SqlServer/Program.cs
```diff
@@ -0,0 +1,55 @@
...
+    public class Program
+    {
+        private string _connectionString;
+        private string _tableName;
+        private SqlQueries _sqlQueries;
+
+        public void Main(string[] args)
+        {
+            //TODO1: use CommandLine and error checks
+            _connectionString = args[0];
```
divega: Is it ok to require customers to provide the whole connection string here or could this be just a configuration key? If we decide to do the latter we should probably find a common way to do it across frameworks (or a convention that we all agree is ok, e.g. looking only at secrets and config.json).

src/Microsoft.Framework.Caching.SqlServer/Program.cs
```diff
@@ -0,0 +1,55 @@
...
+            _tableName = args[1];
```
divega: What if the table exists? Do we want to add some logic to throw a nice exception? or make it a no-op?

src/Microsoft.Framework.Caching.SqlServer/Program.cs
```diff
@@ -0,0 +1,55 @@
...
+            _sqlQueries = new SqlQueries(_tableName);
```
divega: Nit: Maybe just move this field to be a local variable in `CreaateTableAndIndexes()'

src/Microsoft.Framework.Caching.SqlServer/Program.cs
```diff
@@ -0,0 +1,55 @@
...
+            CreateTableAndIndexes();
+        }
+
+        private void CreateTableAndIndexes()
+        {
+            using (var connection = new SqlConnection(_connectionString))
+            {
+                connection.Open();
+
+                var transaction = connection.BeginTransaction(IsolationLevel.ReadCommitted);
```
divega: I think if you do a using here you will save a few lines of code, e.g. you won't need to explicitly Rollback and you won't need the try block.

src/Microsoft.Framework.Caching.SqlServer/Program.cs
```diff
@@ -0,0 +1,55 @@
...
+
+namespace Microsoft.Framework.Caching.SqlServer
+{
+    public class Program
+    {
+        private string _connectionString;
+        private string _tableName;
+        private SqlQueries _sqlQueries;
+
+        public void Main(string[] args)
```
divega: Extra-nit: consider making this command entry point async and cancellable. Not sure it is needed, i.e. if ctrl+c will just kill the process. Under normal conditions creating the table and the index shouldn't take long but if something blocks it may get annoying.

src/Microsoft.Framework.Caching.SqlServer/SqlQueries.cs
```diff
@@ -0,0 +1,74 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Framework.Caching.SqlServer
+{
+    internal class SqlQueries
+    {
+        private const string CreateTableFormat = "CREATE TABLE [{0}](" +
+            "Id nvarchar(100)  NOT NULL PRIMARY KEY, " +
+            "Value varbinary(MAX) NOT NULL, " +
+            "ExpiresAtTimeUTC datetimeoffset NOT NULL, " +
```
divega: This look a bit sillly to me. Not sure if I have all the context, but if it is UTC why use a type that has an offset? Instead we could have datetime2.

divega: Almost orthogonal to that, DATETIME2 and DATETIMEOFFSET are currently not supported in the Mono implementation of SqlClient. I believe that DATETIME2 support is coming sooner.

src/Microsoft.Framework.Caching.SqlServer/SqlQueries.cs
```diff
@@ -0,0 +1,74 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Framework.Caching.SqlServer
+{
+    internal class SqlQueries
+    {
+        private const string CreateTableFormat = "CREATE TABLE [{0}](" +
```
divega: Don't we want to support schemas with the same table name, e.g. for multi-tenancy? In that case you will have to do something about the square brackets, e.g. [schemaname].[tablename] is ok, but I believe that [schemanam.tablename] is not.

src/Microsoft.Framework.Caching.SqlServer/SqlQueries.cs
```diff
@@ -0,0 +1,74 @@
...
+        private const string UpdateCacheItemExpirationFormat = "UPDATE [{0}] SET ExpiresAtTimeUTC = @ExpiresAtTimeUTC " +
+            "WHERE Id = @Id";
+
+        public const string DeleteExpiredCacheItemsFormat = "DELETE FROM [{0}] WHERE @UtcNow > ExpiresAtTimeUTC";
+
+        public SqlQueries(string tableName)
+        {
+            //TODO: error checks
+
+            CreateTable = string.Format(CreateTableFormat, tableName);
```
divega: Is it ok to assume that the user of the command line tool is trusted? I imagine in most situations it better be but otherwise we should probably escape (in the SQL sense) the tableName before we pass it to the `Format()` method. Otherwise it seems it would be pretty easy to get this tool to execute arbitrary SQL.

src/Microsoft.Framework.Caching.SqlServer/SqlQueries.cs
```diff
@@ -0,0 +1,74 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Framework.Caching.SqlServer
+{
+    internal class SqlQueries
+    {
+        private const string CreateTableFormat = "CREATE TABLE [{0}](" +
+            "Id nvarchar(100)  NOT NULL PRIMARY KEY, " +
```
divega: If we don't know how people are going to use it we could increase the size to 900. I think 900 sounds like a big number but I believe there is no downsize to use that.

src/Microsoft.Framework.Caching.SqlServer/SqlQueries.cs
```diff
@@ -0,0 +1,74 @@
...
+            "Value varbinary(MAX) NOT NULL, " +
+            "ExpiresAtTimeUTC datetimeoffset NOT NULL, " +
+            "SlidingExpirationInTicks bigint NULL," +
+            "AbsoluteExpiration datetimeoffset NULL)";
+
+        private const string CreateNonClusteredIndexOnExpirationTimeFormat
+            = "CREATE NONCLUSTERED INDEX Index_ExpiresAtTimeUTC ON [{0}](ExpiresAtTimeUTC)";
+
+        // Used for getting table schema when trying to 'Connect' to the database
+        private const string GetTableSchemaFormat = "SELECT * FROM [{0}]";
```
divega: Nit: you could have the indexes hard-coded in this class, since we own all the queries.

src/Microsoft.Framework.Caching.SqlServer/SqlQueries.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        private const string GetCacheItemFormat =
+            "SELECT Id, Value, ExpiresAtTimeUTC, SlidingExpirationInTicks, AbsoluteExpiration " +
+            "FROM [{0}] WHERE Id = @Id AND @UtcNow <= ExpiresAtTimeUTC";
+
+        private const string AddCacheItemFormat = "INSERT INTO [{0}] " +
```
divega: We can probably come up with a batch that will do `UPSERT` semantics, e.g. only INSERT if is necessary, still make it atomic so that there is no chance of race conditions.

divega: Actually, thinking about it more, if we want to avoid race conditions, and get last-one-wins semantics without blocking we are probably going to need to do something smarter. Not sure yet what.

src/Microsoft.Framework.Caching.SqlServer/SqlServerCache.cs
```diff
@@ -0,0 +1,478 @@
...
+        private int _slidingExpirationInTicksIndex = -1;
+        private int _absoluteExpirationIndex = -1;
+
+        private readonly SqlServerCacheOptions _options;
+        private readonly ILogger _logger;
+        private readonly ISystemClock _systemClock;
+        private readonly SqlQueries _sqlQueries;
+        private DateTimeOffset _lastExpirationScan;
+
+        public SqlServerCache(IOptions<SqlServerCacheOptions> options, ILoggerFactory loggerFactory)
```
divega: Review if we are actually using logging (it is ok if we are going to add it in some places in the near future)

src/Microsoft.Framework.Caching.SqlServer/SqlServerCache.cs
```diff
@@ -0,0 +1,478 @@
...
+            // Try connecting to the database to test if its available and also
+            // get and cache the ordinal positions of the columns for better performance.
+            using (var connection = new SqlConnection(_options.ConnectionString))
+            {
+                var command = new SqlCommand(_sqlQueries.GetTableSchema, connection);
+                connection.Open();
+
+                var reader = command.ExecuteReader(CommandBehavior.SchemaOnly);
+
+                _cacheItemIdIndex = reader.GetOrdinal(CacheItemId);
```
divega: As I said before these could be hardcoded in the SqlQuery class.

src/Microsoft.Framework.Caching.SqlServer/SqlServerCache.cs
```diff
@@ -0,0 +1,478 @@
...
+        public async Task ConnectAsync()
+        {
+            // Try connecting to the database to test if its available and also
+            // get and cache the ordinal positions of the columns for better performance.
+            using (var connection = new SqlConnection(_options.ConnectionString))
+            {
+                var command = new SqlCommand(_sqlQueries.GetTableSchema, connection);
+                await connection.OpenAsync();
+
+                var reader = await command.ExecuteReaderAsync(CommandBehavior.SchemaOnly);
```
divega: Consider specifying sequential access to avoid overhead of filling whole row into memory, in particular in the async path this could be important (the only way to know is testing the perf :smile:). See http://blogs.msdn.com/b/adonet/archive/2012/07/15/using-sqldatareader-s-new-async-methods-in-net-4-5-beta-part-2-examples.aspx.

src/Microsoft.Framework.Caching.SqlServer/SqlServerCache.cs
```diff
@@ -0,0 +1,478 @@
...
+
+            using (var connection = new SqlConnection(_options.ConnectionString))
+            {
+                var command = new SqlCommand(_sqlQueries.GetCacheItem, connection);
+                command.Parameters.AddWithValue(CacheItemId, key);
+                command.Parameters.AddWithValue("UtcNow", utcNow);
+
+                connection.Open();
+
+                var reader = command.ExecuteReader(CommandBehavior.SingleRow);
```
divega: Consider specifying sequential access to avoid overhead of filling whole row into memory, in particular in the async path this could be important (the only way to know is testing the perf :smile:). See http://blogs.msdn.com/b/adonet/archive/2012/07/15/using-sqldatareader-s-new-async-methods-in-net-4-5-beta-part-2-examples.aspx.

src/Microsoft.Framework.Caching.SqlServer/SqlServerCache.cs
```diff
@@ -0,0 +1,478 @@
...
+
+        public byte[] Get([NotNull] string key)
+        {
+            byte[] cacheItemvalue = null;
+            var utcNow = _systemClock.UtcNow;
+
+            using (var connection = new SqlConnection(_options.ConnectionString))
+            {
+                var command = new SqlCommand(_sqlQueries.GetCacheItem, connection);
+                command.Parameters.AddWithValue(CacheItemId, key);
```
divega: Consider specifying the size to avoid cache fragmentation on the server (the size of the parameter is part of the query, we learned this in EF and LINQ to SQL the hard way, but also feel free to perf-test it to make sure that is still the case :smile:)

src/Microsoft.Framework.Caching.SqlServer/SqlServerCache.cs
```diff
@@ -0,0 +1,478 @@
...
+        {
+            var utcNow = _systemClock.UtcNow;
+            var expirationInfo = GetExpirationInfo(utcNow, options);
+
+            using (var connection = new SqlConnection(_options.ConnectionString))
+            {
+                // Try updating the cache item and if it does not succeed then its probably a new item
+                var command = new SqlCommand(_sqlQueries.UpdateCacheItem, connection);
+                command.Parameters.AddWithValue(CacheItemId, key);
+                command.Parameters.AddWithValue(CacheItemValue, value);
```
divega: Consider setting the size of the value parameter as well to avoid query cache fragmentation on the server (feel free to double-check that is has an incidence before you do it). What you can do here is specify an arbitrary size in which the most typical payloads will fit unless the payload is larger than that size.

src/Microsoft.Framework.Caching.SqlServer/SqlServerCache.cs
```diff
@@ -0,0 +1,478 @@
...
+
+                var reader = await command.ExecuteReaderAsync(CommandBehavior.SingleRow);
+
+                DateTimeOffset oldExpirationTime;
+                TimeSpan? slidingExpiration = null;
+                DateTimeOffset? absoluteExpiration = null;
+                if (await reader.ReadAsync())
+                {
+                    oldExpirationTime = await reader.GetFieldValueAsync<DateTimeOffset>(_expiresAtTimeUTCIndex);
+                    cacheItemvalue = await reader.GetFieldValueAsync<byte[]>(_cacheItemValueIndex);
```
divega: Consider using `GetStream()` and `CopyToAsync()` here. This could have more of an impact for the async path.

src/Microsoft.Framework.Caching.SqlServer/SqlServerCache.cs
```diff
@@ -0,0 +1,478 @@
...
+                connection.Open();
+
+                var reader = command.ExecuteReader(CommandBehavior.SingleRow);
+
+                DateTimeOffset oldExpirationTime;
+                TimeSpan? slidingExpiration = null;
+                DateTimeOffset? absoluteExpiration = null;
+                if (reader.Read())
+                {
+                    oldExpirationTime = reader.GetFieldValue<DateTimeOffset>(_expiresAtTimeUTCIndex);
```
divega: Since we have the optimization in place in which we will double the lifetime of the entry, we could also check here if we are getting an already expired item (based on the period that was actually specified, not the doubled one) and decide to return null, plus delete the item.

src/Microsoft.Framework.Caching.SqlServer/SqlServerCache.cs
```diff
@@ -0,0 +1,478 @@
...
+                {
+                    return null;
+                }
+
+                // Check if the current item has a sliding expiration and is going to expire within the
+                // next timeout period(ex:20 minutes), then extend the expiration time.
+                if (slidingExpiration.HasValue &&
+                    utcNow >= (oldExpirationTime - slidingExpiration.Value))
+                {
+                    var newExpirationTime = GetExpirationTime(slidingExpiration, absoluteExpiration, utcNow);
```
divega: An alternative to the optimization that doubles the expiration time is to send the SELECT and UPDATE/DELETEs necessary to maintain the database updated together in one batch. It doesn't get rid of the overhead of executing more than just the SELECT every time you try to get and item, but at least it gets rid of the extra database round trips.

src/Microsoft.Framework.Caching.SqlServer/SqlServerCache.cs
```diff
@@ -0,0 +1,478 @@
...
+                        command = new SqlCommand(_sqlQueries.UpdateCacheItemExpiration, connection);
+                        command.Parameters.AddWithValue(CacheItemId, key);
+                        command.Parameters.AddWithValue(ExpiresAtTimeUTC, newExpirationTime);
+
+                        command.ExecuteNonQuery();
+                    }
+                }
+            }
+
+            StartScanForExpiredCacheItems();
```
divega: This method name isn't very clear to me, at least based on @kichalla's explanation it doesn't do what I thought it would do when I first read it :smile:

src/Microsoft.Framework.Caching.SqlServer/SqlServerCache.cs
```diff
@@ -0,0 +1,478 @@
...
+
+            return cacheItemvalue;
+        }
+
+        public void Refresh([NotNull] string key)
```
divega: Consider giving `Refresh()` and `RefreshAsync()` its own implementation. It seems it shouldn't need to do half of what `Get()`/`GetAsync()` does.

src/Microsoft.Framework.Caching.SqlServer/SqlQueries.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        private const string CreateNonClusteredIndexOnExpirationTimeFormat
+            = "CREATE NONCLUSTERED INDEX Index_ExpiresAtTimeUTC ON [{0}](ExpiresAtTimeUTC)";
+
+        // Used for getting table schema when trying to 'Connect' to the database
+        private const string GetTableSchemaFormat = "SELECT * FROM [{0}]";
+
+        private const string GetCacheItemFormat =
+            "SELECT Id, Value, ExpiresAtTimeUTC, SlidingExpirationInTicks, AbsoluteExpiration " +
+            "FROM [{0}] WHERE Id = @Id AND @UtcNow <= ExpiresAtTimeUTC";
```
divega: String equality in SQL Server will have different semantics to CLR, e.g. depending on the database collation it can be case insensitive and accent sensitive, etc. It will also ignor trailing blanks. Assuming you want to have invariant culture semantics on database lookups and that you want to make this really robust, you have several (non mutually exclusive) options:
1. Specify a collation that satisfies your need on this column on table creation.
2. Always trim trailling spaces or otherwise this if any is detected everywhere the id is taken as an argument.
3. Always encode the string as a byte[]
4. Check again in memory for equality after you got the value (note that this will only help with lookups).

src/Microsoft.Framework.Caching.SqlServer/SqlQueries.cs
divega: Also take into account the memory optimized tables supported on recent versions of SQL Server. I believe it could give the cache a significant boost. The `CREATE TABLE` statement will look different and it might also give you another reason to consider stored procedures, which can be compiled to native and may give you another significant boost (once more, you would need to try and measure the impact). See https://msdn.microsoft.com/en-us/library/dn133079.aspx.

src/Microsoft.Framework.Caching.SqlServer/SqlServerCache.cs
```diff
@@ -0,0 +1,478 @@
...
+
+            StartScanForExpiredCacheItems();
+        }
+
+        // Called by multiple actions to see how long it's been since we last checked for expired items.
+        // If sufficient time has elapsed then a scan is initiated on a background task.
+        private void StartScanForExpiredCacheItems()
+        {
+            var utcNow = _systemClock.UtcNow;
+            if ((utcNow - _lastExpirationScan) > _options.ExpirationScanFrequency)
```
divega: I think we should avoid concurrency issues here with some sort of lock of coordination.

divega: Can the interval be low enough (i.e. the frequency high enough) that you will have more than one worker thread doing this at the same time? It is possibly just a corner case but if there was a simple and nice pattern to guarantee that there is only one doing it at a time I would probably prefer that. Otherwise we should consider validating that this ExpirationScanFrequency has a reasonable value.

src/Microsoft.Framework.Caching.SqlServer/SqlServerCacheExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+using Microsoft.Framework.Internal;
+
+namespace Microsoft.Framework.DependencyInjection
+{
+    public static class SqlServerCacheExtensions
+    {
+        public static IServiceCollection AddSqlServerCache([NotNull] this IServiceCollection services)
+        {
+            services.AddOptions();
+            services.AddSingleton<IDistributedCache, SqlServerCache>();
```
divega: Should this be TryAddX? I guess it depends on whether it is valid to have multiple implementations of the interface registered at the same time.

src/Microsoft.Framework.Caching.SqlServer/SqlServerCacheExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+            return services;
+        }
+
+        public static void ConfigureSqlServerCache(
```
divega: OT: We seem to now have `AddX()` and `ConfigureX()` as separated things everywhere, probably with the only exception of `AddDbContext()` in EF? I just can't remember if/why we decided to have it this way. Right now it feels like collapsing the two methods into one if both exists would be much nicer. @lodejard @HaoK does anyone remember?

src/Microsoft.Framework.Caching.SqlServer/SqlServerCacheOptions.cs
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Framework.Internal;
+using Microsoft.Framework.OptionsModel;
+
+namespace Microsoft.Framework.Caching.SqlServer
+{
+    public class SqlServerCacheOptions : IOptions<SqlServerCacheOptions>
```
divega: @kichalla and I agreed on remove the interface. Should we do it on the in-memory implementation as well?

src/Microsoft.Framework.Caching.SqlServer/SqlServerCacheOptions.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        public ISystemClock SystemClock { get; set; } = new SystemClock();
+
+        public TimeSpan ExpirationScanFrequency { get; set; } = TimeSpan.FromMinutes(10);
```
divega: Naming: when I read frequency I think about something that the higher the value the more often it happens. Should this be interval or something like that?

src/Microsoft.Framework.Caching.SqlServer/SqlServerCache.cs
divega: We are going to try to make this easier to test by having an alternate constructor that takes a SqlConnection. That way the test and the service can be in the same connection and transaction and it is easier to have the tests observe what the APIs are doing to the database, then clean up rolling back the transaction.

src/Microsoft.Framework.Caching.SqlServer/SqlServerCacheExtensions.cs
HaoK: Putting on my @lodejard hat, its because configuring services is seperate from configuring your app.  Its a common pattern to setup your DI in one method (ConfigureServices), and then setup the app in (Configure).  Developers are used to this paradigm.  That said, I think with the new hosting builder changes(IStartupFilter and AutoRequestServices), we have the infrastructure to add functionality which automatically registered a frameworks services and associated middleware in some ordered fashion if we wantd.  Some kind of UseFrameworks() list which contained an ordered list of frameworks who's services and middlewares were added first.

divega:
I believe transaction isolation level is going to have an impact on the perf and scalability of this because of the locking, although it is hard to predict how much. In general I would expect SNAPSHOT and READ COMMITTED with the SNAPSHOT option to be better. We should consider our options, e.g. explicitly wrapping some operations in transactions with SNAPSHOT. Otherwise encourage customers to set READ_COMMITTED_SNAPSHOT themselves on their databases.

test/Microsoft.Framework.Caching.SqlServer.Tests/SqlServerCacheTest.cs
```diff
@@ -0,0 +1,496 @@
...
+            var key = Guid.NewGuid().ToString();
+            var sqlServerCache = new SqlServerCache(options, new LoggerFactory().AddConsole());
+            await sqlServerCache.ConnectAsync();
+            await sqlServerCache.SetAsync(
+                key,
+                Encoding.UTF8.GetBytes("Hello, World!"),
+                new DistributedCacheEntryOptions().SetSlidingExpiration(TimeSpan.FromSeconds(10)));
+
+            // set the clock's UtcNow far in future
+            testClock.Add(TimeSpan.FromHours(10));
```
divega: Wow, I love this!

kichalla:
/cc: @Tratcher @divega : I have updated the PR. Please take a look.

@divega : I haven't addressed all of your comments and would like discuss with you offline, but probably you can take a look at the ones that are addressed.

test/Microsoft.Framework.Caching.SqlServer.Tests/SqlServerCacheTest.cs
```diff
@@ -0,0 +1,574 @@
...
+using Microsoft.Framework.Internal;
+using Microsoft.Framework.Logging;
+using Microsoft.Framework.OptionsModel;
+using Xunit;
+
+namespace Microsoft.Framework.Caching.SqlServer
+{
+    // This requires SQL Server database to be setup
+    // public
+    public class SqlServerCacheTest
```
kichalla: Will disable the test before checking in.

src/Microsoft.Framework.Caching.SqlServer/SqlServerCacheOptions.cs
```diff
@@ -0,0 +1,101 @@
...
+namespace Microsoft.Framework.Caching.SqlServer
+{
+    public class SqlServerCacheOptions
+    {
+        private readonly TimeSpan MinimumExpiredItemsDeletionInterval;
+        private TimeSpan _expiredItemsDeletionInterval;
+
+        public SqlServerCacheOptions()
+        {
+            _expiredItemsDeletionInterval = MinimumExpiredItemsDeletionInterval;
```
Tratcher: You're using MinimumExpiredItemsDeletionInterval before setting it.

src/Microsoft.Framework.Caching.SqlServer/SqlServerCacheOptions.cs
```diff
@@ -0,0 +1,101 @@
...
+                    throw new ArgumentException(
+                        $"{ExpiredItemsDeletionInterval} cannot be less the minimum value of " +
+                        $"{MinimumExpiredItemsDeletionInterval.TotalMinutes} minutes.");
+                }
+
+                _expiredItemsDeletionInterval = value;
+            }
+        }
+
+        private string _connectionString;
```
Tratcher: Organization: fields go at the top of the class.

src/Microsoft.Framework.Caching.SqlServer/SqlServerCacheOptions.cs
kichalla: My bad. Added unit tests now.

kichalla: @Tratcher : update the PR. I removed the checks from here as we actually them in SqlServerCache.cs itself.

src/Microsoft.Framework.Caching.SqlServer/SqlServerCache.cs
```diff
@@ -0,0 +1,166 @@
...
+                    $"value of {MinimumExpiredItemsDeletionInterval.TotalMinutes} minutes.");
+            }
+
+            _systemCock = cacheOptions.SystemClock ?? new SystemClock();
+            _expiredItemsDeletionInterval = cacheOptions.ExpiredItemsDeletionInterval;
+
+            // SqlClient library on Mono doesn't have support for DateTimeOffset and also
+            // it doesn't have support for apis like GetFieldValue, GetFieldValueAsync etc.
+            // So we detect the platform to perform things differently for Mono vs. non-Mono platforms.
+            if (PlatformHelper.IsMono)
```
Tratcher: What about Linux/Mac CoreClr?

src/Microsoft.Framework.Caching.SqlServer/SqlServerCache.cs
```diff
@@ -0,0 +1,166 @@
...
+using Microsoft.Framework.OptionsModel;
+
+namespace Microsoft.Framework.Caching.SqlServer
+{
+    /// <summary>
+    /// Distributed cache implementation using Microsoft SQL Server database.
+    /// </summary>
+    public class SqlServerCache : IDistributedCache
+    {
+        private readonly TimeSpan MinimumExpiredItemsDeletionInterval = TimeSpan.FromHours(1);
```
Tratcher: An hour seems pretty high for a minimum. A server under load could accumulate entries quickly.  I would put it down at 1-5 minutes, with the default at 30-60 mins.

Tratcher: This field should be static.

src/Microsoft.Framework.Caching.SqlServer/SqlServerCacheOptions.cs
```diff
@@ -0,0 +1,36 @@
...
+    {
+        /// <summary>
+        /// An abstraction to represent the clock of a machine in order to enable unit testing.
+        /// </summary>
+        public ISystemClock SystemClock { get; set; }
+
+        /// <summary>
+        /// The periodic interval to scan and delete expired items in the cache. Default is 60 minutes.
+        /// </summary>
+        public TimeSpan ExpiredItemsDeletionInterval { get; set; }
```
Tratcher: You don't appear to be setting a default anywhere, just a minimum.

src/Microsoft.Framework.Caching.SqlServer/project.json
```diff
@@ -0,0 +1,29 @@
+{
+    "commands": {
+        "sqlservercache": "Microsoft.Framework.Caching.SqlServer"
+    },
+    "dependencies": {
+        "Microsoft.Framework.Caching.Abstractions": "1.0.0-*",
+        "Microsoft.Framework.Logging": "1.0.0-*",
+        "Microsoft.Framework.Logging.Abstractions": "1.0.0-*",
```
Tratcher: Logging.Abstracts looks like a redundant dependency.

src/Microsoft.Framework.Caching.SqlServer/DatabaseOperations.cs
```diff
@@ -0,0 +1,377 @@
...
+
+                connection.Open();
+
+                try
+                {
+                    upsertCommand.ExecuteNonQuery();
+                }
+                catch (SqlException)
+                {
+                    // There is a possibility that multiple requests can try to add the same item to the cache, in
```
Tratcher: Is that the only possible error? If so log it as Info or Warning.  If there are other possible errors we shouldn't be masking them.

kichalla: Yeah, I tried to look for an exception type which is specific for 'duplcate key' but couldn't fine one. On searching the web, I see some people using hresult to figure it out and also they seem to be not accurate based on their description. Question: How would logging help though...like what action can a user possibly take here?

src/Microsoft.Framework.Caching.SqlServer/SqlQueries.cs
```diff
@@ -0,0 +1,91 @@
...
+        private const string SetCacheItemFormat =
+            "IF NOT EXISTS(SELECT Id FROM {0} WHERE Id = @Id) " +
+            "BEGIN " +
+                "INSERT INTO {0} " +
+                    "(Id, Value, ExpiresAtTimeUTC, SlidingExpirationInTicks, AbsoluteExpirationUTC) " +
+                    "VALUES (@Id, @Value, @ExpiresAtTimeUTC, @SlidingExpirationInTicks, @AbsoluteExpirationUTC) " +
+            "END " +
+            "ELSE " +
+            "BEGIN " +
+                "UPDATE {0} SET Value = @Value, ExpiresAtTimeUTC = @ExpiresAtTimeUTC " +
```
Tratcher: You should set all fields, they can all change if I call Set twice with different options.

src/Microsoft.Framework.Caching.SqlServer/SqlServerCacheOptions.cs
kichalla: yeah, good catch!

src/Microsoft.Framework.Caching.SqlServer/SqlQueries.cs
kichalla: Will change but feels weird to set different options on different occasions.

src/Microsoft.Framework.Caching.SqlServer/project.json
Eilon: It's redundant, but it's the one that should be kept. The non-Abstractions Logging reference should be removed.

kichalla: Actually I need the non-abstractions logging as I need to instantiate `LoggerFactory` to which the console logger provider is added.

Eilon: Hmm I just noticed that both the command line util and the provider itself seem to be in the same package. Should those not be split out into separate packages? Their functionality probably doesn't overlap a whole lot. And their usage is certainly orthogonal. One is a utility used during deployment, the other is a runtime feature. They're probably never even used on the same machine, except during development.

src/Microsoft.Framework.Caching.SqlServer/project.json
```diff
@@ -0,0 +1,28 @@
...
+    "dependencies": {
+        "Microsoft.Framework.Caching.Abstractions": "1.0.0-*",
+        "Microsoft.Framework.CommandLineUtils.Sources": { "version": "1.0.0-*", "type": "build" },
+        "Microsoft.Framework.Logging": "1.0.0-*",
+        "Microsoft.Framework.Logging.Console": "1.0.0-*",
+        "Microsoft.Framework.NotNullAttribute.Sources": { "type": "build", "version": "1.0.0-*" },
+        "Microsoft.Framework.OptionsModel": "1.0.0-*"
+    },
+    "description": "ASP.NET 5 distributed cache implementation on Microsoft SQL Server.",
+    "frameworks": {
```
PinpointTownes: Out of curiosity, have you considered adding a `net45(1)` target like `Microsoft.Framework.Caching.Abstractions`?

https://github.com/aspnet/Caching/blob/dev/src/Microsoft.Framework.Caching.Abstractions/project.json

src/Microsoft.Framework.Caching.SqlServer/MonoDatabaseOperations.cs
```diff
@@ -0,0 +1,291 @@
...
+
+                connection.Open();
+
+                try
+                {
+                    upsertCommand.ExecuteNonQuery();
+                }
+                catch (SqlException ex)
+                {
+                    if (ex.Message.IndexOf(DuplicateKeyErrorText, StringComparison.OrdinalIgnoreCase) > 0)
```
Tratcher: Never read/parse exception messages, they get localized. Is there any other way to isolate this error?

kichalla: Sure, I updated the PR now where we can check a message number which seem to be the one popularly suggested.

I suppose in the following table the message-id and language-id both combined are unique and so is safe to use this message code for different languages, but i could be wrong as this is not a table but a sql view. @divega 

image

Tratcher:
Looks pretty good. I'll defer to @divega at this point.

src/Microsoft.Framework.Caching.SqlServer/Columns.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Framework.Caching.SqlServer
+{
+    public static class Columns
```
kichalla: From review: make this `internal`

Eilon:
I'm sure my comment was *somewhere* in the thread, but I don't recall seeing a response...

I think we need to split out the *management* commands out from the actual provider. The two are orthogonal concerns and shouldn't be in the same package. I'm not concerned whether that's addressed in this PR or a follow-up PR.

kichalla:
@Eilon My bad, I missed addressing your comment...sure, I will split the packages into 2 ..one with the name `Microsoft.Framework.Caching.SqlServer` having the caching code and the other for management commands with the package name `SqlServerCache` (so that from dnu the commands install/uninstall will consistently use this name)..sounds good?

kichalla:
@Eilon The latest commit in this PR has the project split into 2 as mentioned earlier. Please take a look.

src/Microsoft.Framework.Caching.SqlServer/project.json
kichalla: Thanks @PinpointTownes ..actually I haven't thought about it...

src/SqlServerCache/Program.cs
```diff
@@ -0,0 +1,132 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Data;
+using System.Data.SqlClient;
+using Microsoft.Framework.Logging;
+using Microsoft.Framework.Runtime.Common.CommandLine;
+
+namespace SqlServerCache
```
Tratcher: @Eilon Didn't we decide to change all the commands packages to Microsoft.* rather than having a flat name?

kichalla: :+1: you mean someting like `Microsoft.Commands.SqlServerCache`?

Eilon: Should discuss with engineering team. @kichalla can you start a thread and also include @Tratcher ? I do think we should use long names. These things aren't typed almost ever.

Eilon:
The split looks good to me, thanks!

src/Microsoft.Framework.Caching.SqlServer/SqlQueries.cs
```diff
@@ -0,0 +1,71 @@
...
+    {
+        private const string TableInfoFormat =
+            "SELECT TABLE_CATALOG, TABLE_SCHEMA, TABLE_NAME, TABLE_TYPE " +
+            "FROM INFORMATION_SCHEMA.TABLES " +
+            "WHERE TABLE_SCHEMA = '{0}' " +
+            "AND TABLE_NAME = '{1}'";
+
+        private const string GetCacheItemFormat =
+            "SELECT Id, ExpiresAtTime, SlidingExpirationInTicks, AbsoluteExpiration, Value " +
+            "FROM {0} WHERE Id = @Id AND @UtcNow <= ExpiresAtTime";
```
lodejard: Consider putting back quotes around table name {0} because it is user-provided. http://stackoverflow.com/questions/11321491/when-to-use-single-quotes-double-quotes-and-backticks

src/Microsoft.Framework.Caching.SqlServer/SqlQueries.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        private const string GetCacheItemExpirationInfoFormat =
+            "SELECT Id, ExpiresAtTime, SlidingExpirationInTicks, AbsoluteExpiration " +
+            "FROM {0} WHERE Id = @Id AND @UtcNow <= ExpiresAtTime";
+
+        private const string SetCacheItemFormat =
+            "IF NOT EXISTS(SELECT Id FROM {0} WHERE Id = @Id) " +
```
lodejard: I wonder - could this be changed to an UPDATE statement followed by an INSERT if the `@@ROWCOUNT == 0` afterwards? In that case you'd be dropping to 1 operation instead of 2 for the update case by skipping the exists check.

lodejard:
two small notes on the sql commands, but looks good to me otherwise

src/Microsoft.Framework.Caching.SqlServer/SqlQueries.cs
kichalla: Sure, I have updated the PR now. The quotes are escaped now.

src/Microsoft.Framework.Caching.SqlServer/SqlQueries.cs
kichalla: Good point! I have updated the PR with this change and was able to run all tests successfully.

kichalla:
@Tratcher After @divega  feedback about doing select-update (when we try get a cache item) in a single statement, I have moved all the expiration related logic check to the sql query(so big thanks to @divega ...he helped in building up with the query). So there are considerable about of changes to the design now. So we no longer have expiration time multiplier and honor the idle timeout as expected. Please review the code again.

@divega @Eilon @lodejard I updated the PR. Please take a look.

@Eilon : AppVeyor and Travis fail as i have tests which connect to Sql Server. I will disable them before checking in. Will discuss with you on how to enable this in long run as this requires setup of sql server which complicates our CI story.

Eilon:
@kichalla the updates for the stuff I was looking at look good.

Regarding the Travis stuff - I'm pretty sure we don't want to run stuff that uses SQL Server on these CI's...

MichaCo:
@Eilon @kichalla It should be possible to run the tests on AppVeyor at least. You can start all kind of services if needed: http://www.appveyor.com/docs/services-databases

Same [for Travis](http://docs.travis-ci.com/user/database-setup/), no MS SQL Server of course ;)

Eilon:
@MichaCo oh for sure, good to know. I was more saying that I don't *want* to run them there, not that we can't :smile: My real reason is that these tests would likely be slow, and quite possibly flakey, and our AppVeyor queues tend to get really backed up (even with our private queue), so I'm really favoring speed and reliability here.

Tratcher:
FYI: https://github.com/aspnet/Caching/pull/74

kichalla:
Yeah, I have already made changes to my code to reflect that(latest commit)...but in my case `Task.Run` is sufficient though.

src/Microsoft.Framework.Caching.SqlServer/SqlQueries.cs
```diff
@@ -0,0 +1,96 @@
...
+                "VALUES (@Id, @Value, @ExpiresAtTime, @SlidingExpirationInSeconds, @AbsoluteExpiration); " +
+            "END ";
+
+        private const string DeleteCacheItemFormat = "DELETE FROM {0} WHERE Id = @Id";
+
+        public const string DeleteExpiredCacheItemsFormat = "DELETE FROM {0} WHERE @UtcNow > ExpiresAtTime";
+
+        public SqlQueries(string schemaName, string tableName)
+        {
+            //TODO: sanitize schema and table name
```
Tratcher: Do you plan to file bugs for this TODO?

src/Microsoft.Framework.Caching.SqlServer/project.json
Tratcher: Yes, add it.

Tratcher:
:shipit:

PR #71, opened 2015-06-21
Update sample project's framework dependencies to dotnet.

CatPhat:
Resolves missing dependency issues with the sample project's project.json file.

### DIFF
--- a/samples/MemoryCacheConcurencySample/project.json
+++ b/samples/MemoryCacheConcurencySample/project.json
@@ -4,9 +4,10 @@
     },
     "commands": { "MemoryCacheConcurencySample": "MemoryCacheConcurencySample" },
     "frameworks": {
-        "dnx451": { },
-        "dnxcore50": {
-            "dependencies": {
+      "net45": { },
+      "dnx451": { },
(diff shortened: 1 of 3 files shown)


Tratcher:
@anurse ??

anurse:
Applications, samples included, should not be using `dotnet`.

CatPhat:
Ah sorry, had assumed that ```dotnet```  was now being used due to the references in the other ```Microsoft.Framework.Caching.*``` projects.

anurse:
No problem :). It's not fully documented yet so I can totally understand the confusion! We'll have more detailed docs out soon, but `dotnet` should generally be used in Class Libraries, and not in code that is directly executed (so, Web/Console Apps, Unit Tests, etc.)

Tratcher:
@anurse How did we decide which projects to convert? I see that Caching, Logging, and EF were changed, but not HttpAbstractions, Hosting, or DNX. Or have we just not gotten around to it?

anurse:
Things that are tied to the DNX application model should not be converted either. `dotnet` is for things that would previously have been "portable class libraries"

PR #73, opened 2015-07-06
Remove System.Threading.ThreadPool dependency

bricelam:
It prevented the package from installing on UWP projects.

@rowanmiller @natemcmaster

### DIFF
--- a/src/Microsoft.Framework.Caching.Abstractions/project.json
+++ b/src/Microsoft.Framework.Caching.Abstractions/project.json
@@ -10,7 +10,6 @@
             "dependencies": {
                 "System.Collections": "4.0.10-beta-*",
                 "System.Threading": "4.0.10-beta-*",
-                "System.Threading.ThreadPool": "4.0.10-beta-*",
                 "System.Threading.Tasks": "4.0.10-beta-*"
             }
         }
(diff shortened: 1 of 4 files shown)


Tratcher:
Can you do it without creating closures?

bricelam:
@Tratcher What's the concern? It was passing the state around before using some internal mechanism...

davidfowl:
To avoid the memory allocation.

bricelam:
It's potentially allocating *less* memory than before. QueuUserWorkItem does this:
```C#
var tpcallBack = new QueueUserWorkItemCallback(callBack, state, compressStack, ref stackMark);
```

bricelam:
Which is no different than a closure:
```
class Closure
{
    Action<object> _action;
    object _state;

    public Closure(Action<object> action, object state)
    {
        _action = action;
        _state = state;
    }

    public void Invoke() => _action(_state);
}
```

bricelam:
Updated to use the framework's cached delegate.

bricelam:
Closure removed.

anpete:
:shipit:

justinvp:
`Task.Run` should be used to avoid child tasks from attaching to the parent (or alternatively, use `Task.Factory.StartNew` but specify `DenyChildAttach` ([details](http://blogs.msdn.com/b/pfxteam/archive/2012/09/22/new-taskcreationoptions-and-taskcontinuationoptions-in-net-4-5.aspx)).

Also, the delegates are still being allocated each time the methods are called. The calls should be able to be written in a way such that the C# compiler will cache and reuse a single delegate instance.

I'll submit a PR with these changes.

PR #74, opened 2015-07-07
Avoid unnecessary allocations

justinvp:
- Avoid unnecessary delegate allocations (static cached delegates via C# compiler)
- Avoid unnecessary enumerator allocations when enumerating `List<T>` and `Dictionary<TKey, TValue>`
- Specify `TaskCreationOptions.DenyChildAttach` when calling `Task.Factory.StartNew` to prevent child tasks attaching to the parent task ([details](http://blogs.msdn.com/b/pfxteam/archive/2012/09/22/new-taskcreationoptions-and-taskcontinuationoptions-in-net-4-5.aspx))
- Thread safe delegate invocation (via null-propagation operator)

### DIFF
--- a/src/Microsoft.Framework.Caching.Memory/CacheEntry.cs
+++ b/src/Microsoft.Framework.Caching.Memory/CacheEntry.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace Microsoft.Framework.Caching.Memory
@@ -147,27 +148,30 @@ namespace Microsoft.Framework.Caching.Memory
         // TODO: Ensure a thread safe way to prevent these from being invoked more than once;
         internal void InvokeEvictionCallbacks()
         {
-            var callbacks = PostEvictionCallbacks;
-            PostEvictionCallbacks = null;
-            if (callbacks != null)
+            if (PostEvictionCallbacks != null)
             {
-                Task.Factory.StartNew(InvokeCallbacks, callbacks);
+                Task.Factory.StartNew(state => InvokeCallbacks((CacheEntry)state), this,
+                    CancellationToken.None, TaskCreationOptions.DenyChildAttach, TaskScheduler.Default);
             }
         }
 
-        private void InvokeCallbacks(object state)
+        private static void InvokeCallbacks(CacheEntry entry)
         {
-            var callbackRegistrations = (IList<PostEvictionCallbackRegistration>)state;
+            var callbackRegistrations = entry.PostEvictionCallbacks;
+            entry.PostEvictionCallbacks = null;
+
+            if (callbackRegistrations == null)
+            {
+                return;
+            }
+
             for (int i = 0; i < callbackRegistrations.Count; i++)
             {
                 var registration = callbackRegistrations[i];
 
                 try
                 {
-                    if (registration.EvictionCallback != null)
-                    {
-                        registration.EvictionCallback(Key, Value, EvictionReason, registration.State);
-                    }
+                    registration.EvictionCallback?.Invoke(entry.Key, entry.Value, entry.EvictionReason, registration.State);
                 }
                 catch (Exception)
                 {
--- a/src/Microsoft.Framework.Caching.Memory/MemoryCache.cs
+++ b/src/Microsoft.Framework.Caching.Memory/MemoryCache.cs
@@ -13,7 +13,7 @@ namespace Microsoft.Framework.Caching.Memory
 {
     public class MemoryCache : IMemoryCache
     {
-        private readonly IDictionary<string, CacheEntry> _entries;
+        private readonly Dictionary<string, CacheEntry> _entries;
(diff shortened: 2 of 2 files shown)


Tratcher:
In your other comment you said Task.Run could also be used. Is there any reason you choose one over the other?

justinvp:
> In your other comment you said Task.Run could also be used. Is there any reason you choose one over the other?

I mentioned `Task.Run` because it uses the `DenyChildAttach` option under the covers, but I couldn't actually use `Task.Run` here because it doesn't have an overload that takes the `state` parameter, and we want to use that to avoid the closure.

justinvp:
To clarify: previously `ScanForExpiredItems` was a private method. So I could have used `Task.Run` to call that method. But then the delegate to call `ScanForExiredItems` would be created each time. I could have written code to create the delegate instance once and cache it in a field, but that adds an extra field to the class. So Instead, I changed `ScanForExpiredItems` to be static, and used a lambda to call it (passing the instance as the state), which allows the C# compiler to cache a static instance of the delegate.

justinvp:
To better illustrate what I mean, I could have written it like this:

```c#
public class MemoryCacheAllocating
{
    private void RemoveEntries()
    {
        Task.Run(() => ScanForExpiredItems()); // this will allocate a new Action instance each time
    }
    
    private void ScanForExpiredItems()
    {
    }
}
```

To avoid allocating the `Action` instance each time, I could have manually created it once and cached it like this:

```c#
public class MemoryCacheAllocatingFix
{
    private Action _scan;
    
    private void RemoveEntries()
    {
        if (_scan == null)
            _scan = ScanForExpiredItems;
        
        Task.Run(_scan);
    }
    
    private void ScanForExpiredItems()
    {
    }
}
```

But this requires adding a new instance field. Not a huge deal, but it can be avoided. Instead I opted for the following:

```c#
public class MemoryCache
{
    private void RemoveEntries()
    {
        Task.Factory.StartNew(state => ScanForExpiredItems((MemoryCache)state), this,
            CancellationToken.None, TaskCreationOptions.DenyChildAttach, TaskScheduler.Default);
    }
    
    private static void ScanForExpiredItems(MemoryCache cache)
    {
    }
}
```

Making `ScanForExpiredItems` static and calling it like above allows the C# compiler to lazily create and cache the delegate (in a static field of a private class that the compiler generates). `ScanForExpiredItems` needs to use some of the instance fields, so the `state` parameter is used, passing `this` as the state to avoid a closure allocation. `Task.Run` doesn't have an overload that accepts a `state` parameter, so `Task.Factory.StartNew` is used, passing `this` as the state parameter along with the same options that `Task.Run` uses.

You can see the code that Roslyn generates in the above examples [here](http://tryroslyn.azurewebsites.net/#b:master/f:r/K4Zwlgdg5gBAygTxAFwKYFsDcAoUlaIoYB0AKgBYBOqAhgCb457TxJrplW0PRk0gBrEDmwB6UTADCAe2AAbOjHI0Abqhh1pEdcnJgQAGhgAjYMhhhzNOXOkBjGmhAwaMbQHcYAQTvIwWiwgUGgg7dVo7chg/dFRsAAdgYzkwOxg7OX5nAFkMaUoESRpI1C8be0d8bABvbBh6mHjKMBVHdRVpMEUAJTy1AFEIZGbUEAAKAEo6htqGuZhSfgFibuAIMcmYAF4APngHCAAxfP6AD3iwajoASXZxiYmcOYBfafq3xubWtBgOrv2QsdKGcLldbhh7h9Zg1Xq9sGIJDJ5IplGoYOgQsBrHIEOliuRUIpdOo6Kg5KgoG0EkkUmkMlkYLl0PlCvjSuUHH5oIcwKcah8mi02t5fP4IDAAPogA5PBoCr7Cv49PqoQbDMCjSZQj5zMAAMxgYylB22Wzc8jkU3m1sl0pC2wBRxO50uhPB6GEOrlNoWSxWayNdogjw+rzm8qFPyVjqBINdNzuWrm0PqsPh4ik5GkIB00g0ed0+mpyVS6UyIByeQKRRK/Lmgu+7U6yuZAyGI0hya99UWgmIh2KyBZxDgyBolGQADlUO4xsEfrsY87QW7E2MmSyawSJvPUBMjIXDN35kVQmTMn4tKRpAJUBBiJOtKgjL2BJJqJUtAB5eKXoLEAARO9Cj0BQvGQMdIhfJY4BKOh5FQShANQPUaHkZAQxeD4I0bGB51LaNYMBZd43dcYN2rNk8RKK0ZlDbA4SAAA=).

Tratcher:
From a maintenance perspective I'd rather use the second approach. That's the pattern we already use elsewhere, and MemoryCache is a long-lived object with few instances.

The 3rd approach requires understanding additional concepts (i.e. looking up every time what DenyChildAttach means).

justinvp:
> MemoryCache is a long-lived object with few instances.

What about `CacheEntry`? Presumably not as long-lived (expiration) and many instances.

Tratcher:
Oh, good point.
:shipit:

kichalla:
Thanks @justinvp I have merged your changes now

PR #81, opened 2015-07-21
[Fixes #80] Fix exception rethrow to preserve call stack

### LINKED ISSUE #80: Get rid of throw ex
We'll lose the call stack:

https://github.com/aspnet/Caching/blob/dev/src/Microsoft.Framework.Caching.SqlServer/DatabaseOperations.cs#L216

/cc @kichalla

kichalla:
@davidfowl


Eilon:
:shipit:

Issue #79, opened 2015-07-20
IDistrbutedCache Extension Methods/Helper Methods


RehanSaeed:
The IDistrbutedCache seems pretty basic at the moment compared to the IMemoryCache, as it only deals in byte arrays. Most developers using this interface will be writing some extensions methods for it. 

This [StackOverflow post](http://stackoverflow.com/questions/31458950/is-there-any-guidance-for-caching-patterns-in-asp-net-5) is asking if these methods will eventually be part of ASP.NET Caching or if we should role our own.

The ```BinaryFormatter``` is not available in .NET Core, is it going to be added in future? In the mean time the ```BinaryWriter``` and ```BinaryReader``` seem to be the way to go. I might just submit a pull request this evening.

See the StackOverflow post for the proposed extension methods.

RehanSaeed:
I created a pull request (https://github.com/aspnet/Caching/pull/94) to help provoke a discussion about the best way forward.

muratg:
Closing this based on https://github.com/aspnet/Caching/pull/94#discussion_r39995622

PR #94, opened 2015-08-15
IDistributedCache Extension Methods

RehanSaeed:
I have created three sets of extension methods for IDistributedCache.

1. based on BinaryReader & BinaryWriter. This gives us simple types like int and string.
2. Based on BinaryFormatter (Not available for DNX Core) but provides best performance. Caching is all about performance so it makes sense to use this. Is there a chance we could bring the BinaryFormatter to DNX Core?
3. Using JSON.NET Serializer with BinaryReader & BinaryWriter. This is a workaround for the BinaryFormatter not being available. This requires a dependency on NewtonSoft.Json. I chose to use version 6.0.6 because that is what is being used elsewhere in ASP.NET 5.

I hope to provoke a discussion about what we can do to make using
IDistributedCache easier.

### DIFF
--- a/src/Microsoft.Framework.Caching.Abstractions/project.json
+++ b/src/Microsoft.Framework.Caching.Abstractions/project.json
@@ -2,7 +2,7 @@
     "dependencies": {
       "Microsoft.Framework.NotNullAttribute.Sources": {
         "type": "build",
-        "version": "1.0.0-beta5"
+        "version": "1.0.0-*"
       },
       "Newtonsoft.Json": "6.0.6"
     },
@@ -12,12 +12,15 @@
         "dnx451": { },
         "dotnet": {
             "dependencies": {
-                "System.Collections": "4.0.10-beta-23019",
-                "System.Threading": "4.0.10-beta-23019",
-                "System.Threading.ThreadPool": "4.0.10-beta-23019",
-                "System.Threading.Tasks": "4.0.10-beta-23019"
+                "System.Collections": "4.0.10-beta-*",
+                "System.Threading": "4.0.10-beta-*",
+                "System.Threading.Tasks": "4.0.10-beta-*"
             }
         }
     },
-    "version": "1.0.0-beta5"
+    "version": "1.0.0-*",
+    "repository": {
+        "type": "git",
+        "url": "https://github.com/aspnet/caching"
+    }
 }
--- a/src/Microsoft.Framework.Caching.Abstractions/project.json
+++ b/src/Microsoft.Framework.Caching.Abstractions/project.json
@@ -1,6 +1,10 @@
 {
     "dependencies": {
-        "Microsoft.Framework.NotNullAttribute.Sources": { "type": "build", "version": "1.0.0-*" }
+      "Microsoft.Framework.NotNullAttribute.Sources": {
+        "type": "build",
+        "version": "1.0.0-beta5"
+      },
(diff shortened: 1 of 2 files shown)


RehanSaeed:
See also https://github.com/aspnet/Caching/issues/79.

src/Microsoft.Framework.Caching.Abstractions/project.json
```diff
@@ -1,6 +1,10 @@
 {
     "dependencies": {
-        "Microsoft.Framework.NotNullAttribute.Sources": { "type": "build", "version": "1.0.0-*" }
+      "Microsoft.Framework.NotNullAttribute.Sources": {
+        "type": "build",
+        "version": "1.0.0-*"
+      },
+      "Newtonsoft.Json": "6.0.6"
```
davidfowl: This isn't really a good dependency to have in the abstractions package. These extension methods belong elsewhere

RehanSaeed: Totally agree. I only added the Json.NET reference because the BinaryFormatter was not available on .NET Core and using the Json serializer seemed a decent alternative.

Also, even if we do need to add Json.NET, what would be a good place to add these extension methods? There isn't a common caching implementation project, all the projects are specific implementations e.g. memory, redis, sqlserver.

Tratcher: Stuff like this should go in optional add-on packages. Compare to https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNet.Http.Extensions/SessionExtensions.cs. Maybe M.F.C.Distributed.Serializers or .Extensions?

RehanSaeed: Would a new Microsoft.Framework.Caching.Extensions project do the trick?

davidfowl:
@lodejard What do you think about these methods in an extension project? This is related to https://github.com/aspnet/Session/pull/25. We agreed that application code should have methods that (de)serialize higher level objects like this as there are much better ways to serialize objects than BinaryFormatter/BinaryWriter. 

See https://github.com/aspnet/Session/issues/3

muratg:
@lodejard What do you think?  I agree with @davidfowl this logic should live in the app code instead. 

We should probably not take this one in.

Tratcher:
Simple extensions like this would make sense in a Caching.Extensions package.

src/Microsoft.Framework.Caching.Abstractions/project.json
muratg: Per @Tratcher's comment, M.F.Caching.Extensions would be a better place for this

davidfowl: I don't think we want this in the box. Better to start this off as an extension to see how people use the API.

muratg: @davidfowl Do you mean in a separate repo, or in a new project and package in this repo?

davidfowl: I mean not part of the aspnet org. Something to put in aspnet contrib

muratg:
Thanks @RehanSaeed, closing per @davidfowl's comment: https://github.com/aspnet/Caching/pull/94#discussion_r39995622

PR #95, opened 2015-08-19
Fixing tests to not depend on background thread for cleanup verification

kichalla:
@Tratcher what do you think?

### DIFF
--- a/src/Microsoft.Framework.Caching.Memory/CacheEntry.cs
+++ b/src/Microsoft.Framework.Caching.Memory/CacheEntry.cs
@@ -39,6 +39,9 @@ namespace Microsoft.Framework.Caching.Memory
 
         internal object Value { get; private set; }
 
+        // To enable unit testing
+        internal bool RunCleanupInBackgroundThread { get; set; } = true;
+
         private bool IsExpired { get; set; }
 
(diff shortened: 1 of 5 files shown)


Tratcher:
:cry:

Tratcher:
If it's got to be hacked, at least abstract it out like ISystemClock. That said, I'd prefer not to modify the MemoryCache, just the tests.

src/Microsoft.Framework.Caching.Memory/CacheEntry.cs
```diff
@@ -39,6 +39,9 @@ internal class CacheEntry
 
         internal object Value { get; private set; }
 
+        // To enable unit testing
+        internal bool RunCleanupInBackgroundThread { get; set; } = true;
```
davidfowl: Zomg

muratg:
@kichalla What issue does this one fix?  And are we taking it?

muratg:
@kichalla

kichalla:
@muratg I am taking a look at it now.

kichalla:
Discussed with @Tratcher and closing this PR for now and instead make changes as per the following comment to start with: https://github.com/aspnet/Caching/issues/106#issuecomment-140143178

PR #98, opened 2015-08-24
[Fixes #67] Key for memory cache should be of type object instead of …

### LINKED ISSUE #67: Key for memory cache should be of type object instead of string
There doesn't seem to be a good reason for it to be constrained to string and object could have a number of advantages, e.g.:

- Perf: no need to compute a string based on individual components
- Easier to avoid collisions by having separate key types
- Potentially easier to add (and read) metadata to (from) the key
- Encapsulation: cache key can be made non-public, forcing other components to not sniff directly into the cache but talk to the right layer

cc @lodejard @davidfowl and @anpete

kichalla:
…string

@Tratcher 

@divega I guess you wanted to change the following area of code after this change
https://github.com/aspnet/EntityFramework/blob/dev/src/EntityFramework.Core/Query/CompiledQueryCache.cs#L166-L170

### DIFF
--- a/src/Microsoft.Framework.Caching.Abstractions/IMemoryCache.cs
+++ b/src/Microsoft.Framework.Caching.Abstractions/IMemoryCache.cs
@@ -16,24 +16,24 @@ namespace Microsoft.Framework.Caching.Memory
         /// <summary>
         /// Create or overwrite an entry in the cache.
         /// </summary>
-        /// <param name="key">A string identifying the entry. This is case sensitive.</param>
+        /// <param name="key">An object identifying the entry.</param>
         /// <param name="value">The value to be cached.</param>
         /// <param name="options">The <see cref="MemoryCacheEntryOptions"/>.</param>
         /// <returns>The object that was cached.</returns>
-        object Set(string key, object value, MemoryCacheEntryOptions options);
+        object Set(object key, object value, MemoryCacheEntryOptions options);
 
         /// <summary>
         /// Gets the item associated with this key if present.
         /// </summary>
-        /// <param name="key">A string identifying the requested entry.</param>
+        /// <param name="key">An object identifying the requested entry.</param>
         /// <param name="value">The located value or null.</param>
         /// <returns>True if the key was found.</returns>
-        bool TryGetValue(string key, out object value);
+        bool TryGetValue(object key, out object value);
 
         /// <summary>
         /// Removes the object associated with the given key.
         /// </summary>
-        /// <param name="key">A string identifying the entry. This is case sensitive.</param>
-        void Remove(string key);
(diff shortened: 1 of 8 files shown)


Tratcher:
@davidfowl @muratg @Eilon do we actually want to do this?

davidfowl:
Yes we do.

davidfowl:
:shipit:

test/Microsoft.Framework.Caching.Memory.Tests/MemoryCacheSetAndRemoveTests.cs
```diff
@@ -209,5 +209,29 @@ public void RemoveAndReAddFromCallbackWorks()
...
+        {
+            var cache = CreateCache();
+            var obj = new object();
+            var key = new Person { Id = 10, Name = "Mike" };
+
+            var result = cache.Set(key, obj);
+            Assert.Same(obj, result);
+
+            result = cache.Get(key);
+            Assert.Same(obj, result);
```
divega: Unless I am missing something, this is succeeding merely because you used the same instance of Person to set and retrieve the entry. I don't think it is very representative of what needs to happen in the real world. E.g. Person would need to implement structural equality for `Equals()` and compute `GetHashCode()` based on its properties, so that you can use a different instance of it with the same property values to retrieve the same entry. 

I would recommend you do that explicitly and use different instance for `Set()`, `Get()` and `Remove()` or that you just swap to three anonymous type instances of the same structure and values (and the compiler will implement structural equality for you).

src/Microsoft.Framework.Caching.Abstractions/IMemoryCache.cs
```diff
@@ -16,24 +16,24 @@ public interface IMemoryCache : IDisposable
         /// <summary>
         /// Create or overwrite an entry in the cache.
         /// </summary>
-        /// <param name="key">A string identifying the entry. This is case sensitive.</param>
+        /// <param name="key">An object identifying the entry.</param>
```
divega: Should we remind here that is a good idea to use structurally comparable objects for keys? I guess it depends on the target audience.

test/Microsoft.Framework.Caching.Memory.Tests/MemoryCacheSetAndRemoveTests.cs
anpete: @divega It works because it internally uses a Dictionary<> with the default equality comparer. Sent a PR with a regression test so that we don't break it.

divega: I know it works. I just don't think this test is representative of how it should be used. Thanks for adding the test.

PR #101, opened 2015-09-01
Use earliest compatible dependencies

bricelam:
Enables our packages to work on more platforms with more libraries

Note, `AsyncLocal<T>` is the only thing preventing this from working on win8+wp80+wpa81.


bricelam:
DNX isn't handling this well. Will reconsider later.

PR #115, opened 2015-09-30
Downgrade to RTM packages for UWP

bricelam:


Tratcher:
Why does it need more dependencies?

@pranavkm @anurse

bricelam:
@Tratcher It doesn't. The method I use just lists transitive dependencies too, and I didn't bother removing them.

Eilon:
:shipit:

Issue #106, opened 2015-09-13
Need to improve extremely slow (1min) long "unit" tests


Eilon:
These two are super slow to run:

1. **30sec** [GcNotificationTests.CallbackInvokedMultipleTimes](https://github.com/aspnet/Caching/blob/dev/test/Microsoft.Framework.Caching.Memory.Tests/GcNotificationTests.cs#L28-L59)
2. **1min** [MemoryCacheSetAndRemoveTests.SetOverwritesAndInvokesCallbacks](https://github.com/aspnet/Caching/blob/dev/test/Microsoft.Framework.Caching.Memory.Tests/MemoryCacheSetAndRemoveTests.cs#L85-L131)

These take way, waaay too long.

Tratcher:
FYI: These kept getting lengthened because slow CI machines would cause timeouts for the positive tests (Assert.True). The negative tests (Assert.False) don't actually need timeouts that long.

Also https://github.com/aspnet/Caching/pull/95

muratg:
See https://github.com/aspnet/Caching/pull/95/files

muratg:
Putting this in RC1 for now.

muratg:
@BrennanConroy Please adjust the timeouts on the negative tests

BrennanConroy:
Looks like the GCNotificationTest was fixed by #120

BrennanConroy:
834385ee7495571550b2f0eb2ffd2c7fe6725dbe

PR #123, opened 2015-11-18
Fix slow test

BrennanConroy:
#106 

@Tratcher @kichalla


test/Microsoft.Extensions.Caching.Memory.Tests/MemoryCacheSetAndRemoveTests.cs
```diff
@@ -121,13 +121,13 @@ public void SetOverwritesAndInvokesCallbacks()
             });
             result = cache.Set(key, value2, options2);
             Assert.Same(value2, result);
-            Assert.True(callback1Invoked.WaitOne(TimeSpan.FromSeconds(30)), "Callback1");
-            Assert.False(callback2Invoked.WaitOne(TimeSpan.FromSeconds(30)), "Callback2");
+            Assert.True(callback1Invoked.WaitOne(TimeSpan.FromSeconds(5)), "Callback1");
```
Tratcher: The Assert.True timeouts need to be high so they don't timeout / become flaky on slow machines. Only lower the Assert.False timeouts to 1s.

BrennanConroy:
:up: :date:

Tratcher:
:shipit:

Issue #17, opened 2014-09-16
Harden trigger and callback registrations in CacheEntry


Tratcher:
There are possible race conditions in CacheEntry between AttachTriggers and DetatchTriggers. Similar for the callback APIs.  These need to be hardened to prevent potential stress/race crashes.

BrennanConroy:
Every call to InvokeEvictionCallbacks is wrapped in the ReaderWriterLock in MemoryCache so it looks impossible to invoke more than once

BrennanConroy:
#126 and #132

PR #132, opened 2015-12-16
Thread safe way of not invoking Eviction callbacks more than once

BrennanConroy:
Last part of #17
@Tratcher

### DIFF
--- a/src/Microsoft.Extensions.Caching.Memory/CacheEntry.cs
+++ b/src/Microsoft.Extensions.Caching.Memory/CacheEntry.cs
@@ -16,6 +16,8 @@ namespace Microsoft.Extensions.Caching.Memory
 
         private readonly DateTimeOffset? _absoluteExpiration;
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.Extensions.Caching.Memory/CacheEntry.cs
```diff
@@ -47,7 +47,7 @@ internal class CacheEntry
 
         internal IList<IDisposable> ExpirationTokenRegistrations { get; set; }
 
-        internal IList<PostEvictionCallbackRegistration> PostEvictionCallbacks { get; set; }
+        internal IList<PostEvictionCallbackRegistration> PostEvictionCallbacks;
```
Tratcher: if it's a field now it should be _postEvictionCallbacks and listed before the constructor. It should also be private. If anyone outside of this class needs it then make a property wrapper.

BrennanConroy:
:up: :date:

Tratcher:
:shipit:

PR #136, opened 2015-12-28
Removed Connect method from IDistributedCache

kichalla:
This change is going to need changes in other repos, for example Session which calls the `Connect` method.


kichalla:
cc @lodejard @muratg @Tratcher

muratg:
LGTM

Tratcher:
Will you also be removing ISessionStore.Connect? This will move the cache initialization from startup (https://github.com/aspnet/Session/blob/e91ce99c70614efab26825f685dd026e721b0fa0/src/Microsoft.AspNet.Session/SessionMiddleware.cs#L65) to first access and make issues harder to debug.

Tratcher:
Rebased and merged.

PR #142, opened 2016-01-12
Disable test that fails on Mono.

cesarbs:
cc @Tratcher @JunTaoLuo

### DIFF
--- a/test/Microsoft.Extensions.Caching.Memory.Tests/GcNotificationTests.cs
+++ b/test/Microsoft.Extensions.Caching.Memory.Tests/GcNotificationTests.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Threading;
+using Microsoft.AspNet.Testing.xunit;
 using Microsoft.Extensions.Internal;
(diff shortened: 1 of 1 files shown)


Tratcher:
Mono or CoreClr?

JunTaoLuo:
I think the test should be disabled on Mono only since it's only failing on travis. The Mac CoreCLR tests are all passing.

cesarbs:
@JunTaoLuo Yeah, I had to make a choice. I went with OS X since it only fails on Mono on OS X. But disabling it on Mono makes more sense, otherwise we lose CoreCLR coverage on Mac.

cesarbs:
What's really weird about this is that the other test in the same class, which relies on the exact same behavior (i.e. a finalizer being called in order to invoke a callback), passes. Beats me.

cesarbs:
Disabled on Mono.

Tratcher:
:shipit:

JunTaoLuo:
:shipit: Maybe file an issue to fix this in the future?

PR #156, opened 2016-02-25
Make ServiceCollectionExtensions consistent

javiercn:


javiercn:
Things to pay attention to:
* Comments (There's been a lot of copy paste, so mistakes can happen).
* Method names
* Null checks for arguments
* Overload with configuration action calls parameterless overload.
* Parameterless overload does the work and brings in direct depenendencies
  (Like routing on MVC, for most repositories it means making sure AddOptions is called).

test/Microsoft.Extensions.Caching.Memory.Tests/CacheServiceExtensionsTests.cs
```diff
@@ -20,7 +20,7 @@ public void AddCaching_RegistersMemoryCacheAsSingleton()
             var services = new ServiceCollection();
 
             // Act
-            services.AddCaching();
+            services.AddMemoryCache(options => { });
```
Eilon: Why not call the parameterless ctor?

src/Microsoft.Extensions.Caching.Memory/MemoryCacheServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,53 @@
...
+        /// <param name="services">The <see cref="IServiceCollection" /> to add services to.</param>
+        public static void AddMemoryCache(this IServiceCollection services)
+        {
+            if (services == null)
+            {
+                throw new ArgumentNullException(nameof(services));
+            }
+
+            services.AddOptions();
+            services.TryAdd(ServiceDescriptor.Transient<IDistributedCache, LocalCache>());
```
Eilon: Hmm I think I now get why it was called `AddCaching` - it adds both a memory cache and a distributed cache. As such, I actually think that either this new name is wrong, or that it shouldn't add both types of caches. I'm thinking right now that the name is what's fine, and that it shouldn't add both types of caches.

Perhaps is should be:
- `AddMemoryCache` to add *just* `IMemoryCache` --> `MemoryCache`. The doc comment for this is simple.
- `AddMemoryDistributedCache` to add *just* `IDistributedCache` --> `LocalCache`. The doc comment for this needs to say that it's not *really* distributed, but rather a local in-memory implementation.

cc @Tratcher @lodejard

Eilon:
:clock1:

src/Microsoft.Extensions.Caching.Memory/MemoryCacheServiceCollectionExtensions.cs
javiercn: @Eilon Spoke to @lodejard about this. We've kept the change to AddMemoryCache and we've changed the implmentation of AddRedisCache and AddSqlServerCache to override the distributed cache in place set up by AddMemoryCache.

The rationale behind it is that AddRedisCache and AddSqlServerCache are methods that are very specifically tied up to a given implementation (Compared to AddMvc or AddDataProtection which add a bunch of abstractions) so it's ok for them to unilaterally override the distributed cache service with their implementation.

The user experience is as follows. If a user or framework needs caching it will call AddMemoryCache() which will setup the local and distributed memory caches if no cache is present.

Then if a user wants to use Redis or SQL Server as the distributed cache, they can do so by calling AddRedisCache or AddSQLServerCache.

The order in which the methods are called doesn't matter,
AddRedisCache followed by AddMemoryCache will produce the same result as AddMemoryCache folloewed by AddRedisCache.

javiercn:
updated /cc @Eilon

src/Microsoft.Extensions.Caching.Redis/RedisCacheServiceCollectionExtensions.cs
```diff
@@ -9,26 +9,30 @@
...
+    public static class RedisCacheServiceCollectionExtensions
     {
         /// <summary>
         /// Adds Redis distributed caching services to the specified <see cref="IServiceCollection" />.
         /// </summary>
         /// <param name="services">The <see cref="IServiceCollection" /> to add services to.</param>
-        /// <returns>A reference to this instance after the operation has completed.</returns> 
-        public static IServiceCollection AddRedisCache(this IServiceCollection services)
+        /// <param name="setupAction">An action delegate to configure the provided <see cref="RedisCacheOptions"/>.</param>
+        public static void AddRedisCache(this IServiceCollection services, Action<RedisCacheOptions> setupAction)
```
Tratcher: When did chaining become discouraged?

Eilon: @Tratcher we wanted to have a non-void return type for the cases where there was an especially interesting return type, e.g. MVC's or EF's framework-specific builder interface. Just chaining IServiceCollection isn't terribly interesting.

test/Microsoft.Extensions.Caching.SqlServer.Tests/SqlServerCacheServicesExtensionsTest.cs
```diff
@@ -26,5 +29,87 @@ public void AddSqlServerCache_AddsAsSingleRegistrationService()
...
+            var serviceProvider = services.BuildServiceProvider();
+
+            var distributedCache = services.FirstOrDefault(desc => desc.ServiceType == typeof(IDistributedCache));
+
+            Assert.NotNull(distributedCache);
+            Assert.Equal(ServiceLifetime.Scoped, distributedCache.Lifetime);
+            Assert.IsType<SqlServerCache>(serviceProvider.GetRequiredService<IDistributedCache>());
+        }
+
+        private class TestDistributedCache : IDistributedCache
```
Eilon: Can just use a `Mock<IDC>` for this, no?

src/Microsoft.Extensions.Caching.Memory/MemoryCacheServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,53 @@
...
+        /// <param name="services">The <see cref="IServiceCollection" /> to add services to.</param>
+        public static void AddMemoryCache(this IServiceCollection services)
+        {
+            if (services == null)
+            {
+                throw new ArgumentNullException(nameof(services));
+            }
+
+            services.AddOptions();
+            services.TryAdd(ServiceDescriptor.Transient<IDistributedCache, MemoryDistributedCache>());
```
Eilon: This still seems wrong to me. There's no easy way to add just an implementation of MemoryCache without MemoryDistributedCache? Just because both happen to store things in memory that doesn't seem like they both belong here. What if we add more types of caches in the future? Does this method always register all of them with defaults?

javiercn: Memory cache and MemoryDistributedCache are always required for our system to work as far as I understood. That's the reason why they are bundled together and not done as two separate calls? (The other option we contemplated). I really want to avoid making the user make two calls to extension methods on IServiceCollection to configure the system correctly (it's very easy to forget one).

Another option, although @lodejard doesn't like it very much, it's to have an ICachingBuilder and make AddRedisCache and AddSqlServerCache handle from there. That way a user would do something like

```
services
    .AddCaching()
    .UseRedisCache(ro => {})
```

As far as I understood, what @lodejard doesn't like about this pattern is that we are technically not extending the system but changing how it works. Maybe you feel different about it.

In any case, the other changes (Make AddRedisCache and AddSqlServerCache call Add instead of TryAdd) are still useful as they take away the order dependency on how you call the methods.
(AddCaching().AddRedisCache() would not work)

Also, caching is one of those things I would resolve with a factory. If you need different caches for different parts of the system, a factory is more flexible that just using pure DI. You can inject a factory wherever your cache is needed and the factory can have a create method with parameters that you can use to pick the cache type most appropriate to your component.

So after all this, I think we have 4 options:
* Keep everything as it is right now.
* Use what I have in the current PR.
* Use the builder pattern I proposed as alternative.
* Split AddCaching into AddLocalCache() and AddMemoryCache() (we can add the DistributedSuffix to Add(Memory|Redis|SqlServer)Cache.

In my opinion 2 and 3 are ok.

@Eilon what is your take?

Tratcher: What components require IDistributedCache? Session is the only one I can find.

javiercn:
@sebastienros Doesn't your TagHelper cache abstraction require a distributed cache too?

src/Microsoft.Extensions.Caching.Memory/MemoryCacheServiceCollectionExtensions.cs
Eilon: First, what @Tratcher said. Second, just because we happen to have certain components with certain requirements doesn't make them quite such broad concepts. In-memory caching and distributing caching are two completely different features. In some sense they shouldn't even be in the same abstractions package, but meh.

Eilon: As far as which options to go with: I think these are all completely separate `services.AddXyzCache()` methods. I don't think the builder pattern makes sense here because there's no concrete relationship between these services (we just happened to implement one using the other, but that's an implementation choice, not a contract obligation).

sebastienros:
Yes it requires one, this is why I will be waiting for this PR.

test/Microsoft.Extensions.Caching.SqlServer.Tests/SqlServerCacheServicesExtensionsTest.cs
javiercn: I don't think so, we are basically resolving this services on the test. I don't know for sure if you could register a mock as an implementation for an interface on DI but It will make the test way more obscure.

Eilon: That's exactly what mocks do and what they're for.

javiercn:
@Eilon Am I good to go here?

test/Microsoft.Extensions.Caching.Redis.Tests/project.json
```diff
@@ -10,7 +10,8 @@
   "frameworks": {
     "net451": {
       "dependencies": {
-        "xunit.runner.console": "2.1.0"
+        "xunit.runner.console": "2.1.0",
+        "Moq": "4.2.1312.1622"
```
Eilon: Sort.

src/Microsoft.Extensions.Caching.Memory/MemoryCacheServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,80 @@
...
+        /// Adds a default implementation of <see cref="IDistributedCache"/> that stores items in memory
+        /// to the <see cref="IServiceCollection" />. Frameworks that require a distributed cache to work
+        /// can safely add this dependency as part of their dependency list to ensure that there is at least
+        /// one implementation available. Users can choose a diferent distributed cache based on SQL Server or
+        /// Redis that will replace this implementation.
+        /// </summary>
+        /// <remarks>
+        /// <see cref="AddDistributedMemoryCache(IServiceCollection)"/> should only be used in single
+        /// server scenarios as this cache stores items in memory and doesn't expand across multiple machines.
+        /// For those scenarios is recommended to use a cache backed by SQL Server or Redis.
```
Eilon: `it is recommended`

src/Microsoft.Extensions.Caching.Memory/MemoryCacheServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,80 @@
...
+
+            services.AddMemoryCache();
+            services.Configure(setupAction);
+        }
+
+        /// <summary>
+        /// Adds a default implementation of <see cref="IDistributedCache"/> that stores items in memory
+        /// to the <see cref="IServiceCollection" />. Frameworks that require a distributed cache to work
+        /// can safely add this dependency as part of their dependency list to ensure that there is at least
+        /// one implementation available. Users can choose a diferent distributed cache based on SQL Server or
```
Eilon: I don't think I'd mention SQL Server and Redis specifically.

Also, typo in `different`.

I would actually just remove this entire last sentence.

src/Microsoft.Extensions.Caching.Memory/MemoryCacheServiceCollectionExtensions.cs
Eilon: Again, wouldn't mention specific implementations, but perhaps keep a mention of using a proper distributed cache.

Eilon:
Almost good :smile: 

:watch:

javiercn:
updated /cc: @Eilon

Eilon:
:shipit:

PR #157, opened 2016-02-26
Caching => new Options pattern

HaoK:
cc @divega @lodejard @davidfowl

### DIFF
--- a/src/Microsoft.Extensions.Caching.Memory/MemoryCacheOptions.cs
+++ b/src/Microsoft.Extensions.Caching.Memory/MemoryCacheOptions.cs
@@ -2,22 +2,29 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using Microsoft.Extensions.Internal;
 using Microsoft.Extensions.Options;
 
 namespace Microsoft.Extensions.Caching.Memory
 {
-    public class MemoryCacheOptions : IOptions<MemoryCacheOptions>
+    public class MemoryCacheOptions
     {
+        public MemoryCacheOptions(IEnumerable<IConfigureOptions<MemoryCacheOptions>> configureOptions = null)
+        {
+            if (configureOptions != null)
+            {
+                foreach (var configure in configureOptions)
+                {
+                    configure.Configure(this);
+                }
+            }
+        }
+
         public ISystemClock Clock { get; set; }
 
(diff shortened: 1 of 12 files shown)


src/Microsoft.Extensions.Caching.Memory/MemoryCacheOptions.cs
```diff
@@ -2,22 +2,29 @@
...
+using System.Collections.Generic;
 using Microsoft.Extensions.Internal;
 using Microsoft.Extensions.Options;
 
 namespace Microsoft.Extensions.Caching.Memory
 {
-    public class MemoryCacheOptions : IOptions<MemoryCacheOptions>
+    public class MemoryCacheOptions
     {
+        public MemoryCacheOptions(IEnumerable<IConfigureOptions<MemoryCacheOptions>> configureOptions = null)
```
divega: I thought you were going to try having a base class with the boilerplate constructor. Did you try? I would ok either way, but it would be good to see how it looks like.

muratg:
IOptions gone again? :)

HaoK:
It should be fatal this time :)

src/Microsoft.Extensions.Caching.Memory/MemoryCacheOptions.cs
HaoK: Wasn't immediately clear how we could get it type correctly, but I didn't spend much time on it, I only tried the following before moving on for now

```
public abstract class Options<TOptions> where TOptions : class
{
    public Options(IEnumerable<IConfigureOptions<TOptions>> configure = null) {
        if (configure != null) foreach (var c in configure) 
             // BLOWS UP: because this is not a TOptions
             c.Configure(this);
    }
}

File: src/Microsoft.Extensions.Caching.Redis/RedisCache.cs [hunk 2]
```diff
@@ -34,14 +33,14 @@ public class RedisCache : IDistributedCache, IDisposable
...
         {
-            if (optionsAccessor == null)
+            if (options == null)
             {
-                throw new ArgumentNullException(nameof(optionsAccessor));
+                throw new ArgumentNullException(nameof(options));
             }
 
-            _options = optionsAccessor.Value;
+            _options = options;
```
Tratcher: Hmm, the redis constructor doesn't validate any of the required arguments like the sql one does. This makes the "You failed to specify RedisCacheOptions" mistake harder to detect.

PR #159, opened 2016-03-03
Target minimal TFMs

pranavkm:


src/Microsoft.Extensions.Caching.Memory/project.json
```diff
@@ -23,7 +23,11 @@
     "Microsoft.Extensions.Options": "1.0.0-*"
   },
   "frameworks": {
-    "net451": {},
+    "net451": {
```
kichalla: Curious why do we need this as netstandard 1.3 includes net451 as per this doc: https://github.com/dotnet/corefx/blob/master/Documentation/architecture/net-platform-standard.md

pranavkm: netstandard1.3 targets net46.

kichalla: Ah ok. So should this target `netstandard1.2` as this can work for net451 and also should work for net46 as it supports `netstandard1.3`, which is a super-set?

kichalla: Thinking more I suppose you are being forced to use 1.3 as the dependencies require it. You can ignore my comment.

PR #163, opened 2016-03-11
Return IServiceCollection from builder methods so that additional calls can be chained

ajcvickers:
Previously the AddCaching method did this. It is also common in other builder methods like this. It doesn't do any harm for those that don't want to chain and allows cleaner code with fewer variable declarations for those that do.


Tratcher:
@Eilon @javiercn This support was intentionally removed, correct?

Tratcher:
https://github.com/aspnet/Caching/pull/156#discussion_r54686083

Eilon:
@Tratcher indeed this was a deliberate change, but we're changing our minds on this and going back to allow chaining.

muratg:
:shipit:

PR #166, opened 2016-03-16
Adding ambient link scopes chain implicitly

sebastienros:
- Any call to `Set` will automatically be enlisted in the ambient entry link.
- Options are dispatched to all parent entry links when an entry is added such that they can be shared.
- A `CacheEntry` can now be associated to a `CacheEntlyLink` so that checking for expiration will also check the expiration (time and tokens) on the entry link's tokens.
- Unit tests ensure that sibling scopes are not affected by expired tokens.

### DIFF
--- a/test/Microsoft.Extensions.Caching.Memory.Tests/EntryLinkExpirationTests.cs
+++ b/test/Microsoft.Extensions.Caching.Memory.Tests/EntryLinkExpirationTests.cs
@@ -80,7 +80,7 @@ namespace Microsoft.Extensions.Caching.Memory
             {
                 cache.Set(key, obj, new MemoryCacheEntryOptions().AddExpirationToken(expirationToken));
 
-                cache.Set(key1, obj, new MemoryCacheEntryOptions().AddEntryLink(link));
+                cache.Set(key1, obj, new MemoryCacheEntryOptions());
             }
 
             Assert.Same(obj, cache.Get(key));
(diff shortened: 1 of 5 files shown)


sebastienros:
/cc 
@lodejard for design review
@pranavkm for grammar and indentation

test/Microsoft.Extensions.Caching.Memory.Tests/EntryLinkExpirationTests.cs
```diff
@@ -94,6 +94,72 @@ public void TokenExpires_LinkedEntry()
         }
 
         [Fact]
+        public void TokenExpires_ParentScopeEntry()
+        {
+            var cache = CreateCache();
+            var obj = new object();
+            string key = "myKey";
```
pranavkm: `var key = "myKey";`

sebastienros: Yeah, but if I do that then I have to change all unit tests in the whole package for consistency ;)

test/Microsoft.Extensions.Caching.Memory.Tests/EntryLinkExpirationTests.cs
```diff
@@ -94,6 +94,72 @@ public void TokenExpires_LinkedEntry()
...
+            }
+
+            Assert.Same(obj, cache.Get(key));
+            Assert.Same(obj, cache.Get(key1));
+
+            expirationToken.Fire();
+
+            object value;
+            Assert.False(cache.TryGetValue(key1, out value));
+            Assert.False(cache.TryGetValue(key, out value));
```
Tratcher: :-1: This design is strange. `key` does not logically depend on `key1` here.

test/Microsoft.Extensions.Caching.Memory.Tests/EntryLinkExpirationTests.cs
```diff
@@ -94,6 +94,72 @@ public void TokenExpires_LinkedEntry()
...
+            Assert.Same(obj, cache.Get(key));
+            Assert.Same(obj, cache.Get(key1));
+            Assert.Same(obj, cache.Get(key2));
+
+            expirationToken.Fire();
+
+            object value;
+            Assert.False(cache.TryGetValue(key1, out value));
+            Assert.False(cache.TryGetValue(key, out value));
+            Assert.True(cache.TryGetValue(key2, out value));
```
Tratcher: ???

Tratcher:
Do you have usage scenarios you can share where this design makes sense? I don't get it.

test/Microsoft.Extensions.Caching.Memory.Tests/EntryLinkExpirationTests.cs
sebastienros: Talked with @Tratcher offline, the open question resides in if we want to handle the case where `key` is set before `key1` in this case, or we only need to support when `key` is set after the inner scopes are disposed. If we just care about the later then we don't need the `_entryLink` in the `CacheEntry`.

sebastienros:
Closing as we are making changes bigger ...

PR #168, opened 2016-03-22
Refactoring API

sebastienros:
- Ambient cache entries are automatically nested
- Removing `IEntryLink`
- Fixing nested cache entries invalidation
- Exposing factory helper methods

Requires @Tratcher, @lodejard


sebastienros:
:up::date:

src/Microsoft.Extensions.Caching.Abstractions/ICacheEntry.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// Sets the value of the cache entry.
+        /// </summary>
+        /// <param name="value"></param>
+        ICacheEntry SetValue(object value);
+
+        /// <summary>
+        /// Sets the priority for keeping the cache entry in the cache during a memory pressure tokened cleanup.
+        /// </summary>
+        /// <param name="priority"></param>
+        ICacheEntry SetPriority(CacheItemPriority priority);
```
Tratcher: This SetFoo, return ICacheEntry style seems out of place on the base interface. We usually keep the interfaces simple and add sugar like this as extensions. I'll keep reviewing and see how it works out.

src/Microsoft.Extensions.Caching.Abstractions/MemoryCacheExtensions.cs
```diff
@@ -21,34 +25,90 @@ public static TItem Get<TItem>(this IMemoryCache cache, object key)
...
 
-        public static object Set(this IMemoryCache cache, object key, object value, MemoryCacheEntryOptions options)
+        public static TItem Set<TItem>(this IMemoryCache cache, object key, TItem value, MemoryCacheEntryOptions options)
         {
-            return cache.Set(key, value, options);
+            using (var entry = cache.CreateEntry(key))
+            {
+                if (options != null)
+                {
+                    if (options.AbsoluteExpiration.HasValue)
```
Tratcher: MemoryCacheEntryOptions and ICacheEntry need to be reconciled. The two should work mostly the same.

src/Microsoft.Extensions.Caching.Abstractions/MemoryCacheExtensions.cs
```diff
@@ -21,34 +25,90 @@ public static TItem Get<TItem>(this IMemoryCache cache, object key)
...
+                }
+
+                entry.SetValue(value);
+            }
+
+            return value;
         }
 
-        public static TItem Set<TItem>(this IMemoryCache cache, object key, TItem value)
+        public static TItem GetOrCreate<TItem>(this IMemoryCache cache, object key, Func<ICacheEntry, TItem> factory)
```
Tratcher: We removed factories like this over concerns that they were misleadingly similar to ConcurrentDictionary.GetOrCreate but didn't have it's locking. @Eilon?

src/Microsoft.Extensions.Caching.Memory/CacheEntry.cs
```diff
@@ -1,47 +1,147 @@
...
+        internal IList<IChangeToken> _expirationTokens;
+        internal DateTimeOffset? _absoluteExpiration;
+        internal TimeSpan? _absoluteExpirationRelativeToNow;
+        internal TimeSpan? _slidingExpiration;
+        internal CacheItemPriority _priority;
 
         internal readonly object _lock = new object();
 
-        internal CacheEntry(
+        public CacheEntry(
```
Tratcher: internal

src/Microsoft.Extensions.Caching.Memory/CacheEntry.cs
```diff
@@ -50,6 +150,23 @@ internal class CacheEntry
...
+                if (!_added)
+                {
+                    _notifyCacheEntryDisposed(this);
+                }
+            }
+            finally
+            {
+                CacheEntryHelper.LeaveScope();
+                PropageOptions(CacheEntryHelper.Current);
+                _added = true;
```
Tratcher: Set _added on line 159.

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -76,67 +76,69 @@ public object Set(object key, object value, MemoryCacheEntryOptions cacheEntryOp
...
 
             _entryLock.EnterWriteLock();
             try
             {
-                if (_entries.TryGetValue(key, out priorEntry))
+                if (_entries.TryGetValue(entry.Key, out priorEntry))
                 {
-                    _entries.Remove(key);
+                    _entries.Remove(entry.Key);
                     priorEntry.SetExpired(EvictionReason.Replaced);
```
Tratcher: Removing the existing entry should happen in SetEntry where you'll only need to take the write lock once.

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -302,6 +258,68 @@ private void EntryExpired(CacheEntry entry)
...
+            DateTimeOffset? absoluteExpiration = null;
+            if (entry._absoluteExpirationRelativeToNow.HasValue)
+            {
+                absoluteExpiration = utcNow + entry._absoluteExpirationRelativeToNow;
+            }
+            else if (entry._absoluteExpiration.HasValue)
+            {
+                if (entry._absoluteExpiration <= utcNow)
+                {
+                    throw new ArgumentOutOfRangeException(
```
Tratcher: SetEntry is only called from Dispose which should never throw. If you remove this I think CheckExpired will handle it as already expired.

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -456,24 +474,6 @@ private void ExpirePriorityBucket(int removalCountTarget, List<CacheEntry> expir
             }
         }
 
-        public void Dispose()
-        {
-            Dispose(true);
-        }
-
-        protected virtual void Dispose(bool disposing)
```
Tratcher: Why were these Dispose methods removed? We do actually need them.

src/Microsoft.Extensions.Caching.Abstractions/MemoryCacheExtensions.cs
Eilon: At this point, what are the differences between them that we want to ensure we avoid conflating?

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
sebastienros: I have no idea! I must have let my machine unlocked too long ... I will add a unit test too.

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
sebastienros: We have a unit test that ensures an exception is thrown when adding an absolute expiration in the past, this case. Dispose actually throws in this case and fulfills the unit test as the SetEntry defines the utcNow to take into account, so I would preferably left it here. We can test the value when it's set as the current time is not valid at that moment.

Tratcher: No, Dispose should never throw. Remove it and update the test to verify the value isn't added to the cache.

src/Microsoft.Extensions.Caching.Abstractions/MemoryCacheExtensions.cs
Tratcher: This code should be simpler now with your updated ICacheEntry

src/Microsoft.Extensions.Caching.Abstractions/MemoryCacheExtensions.cs
Tratcher: I recall the concern was around ConcurrentDictionary locking for the duration of the factory and only invoking it once, but the docs don't actually say that. https://msdn.microsoft.com/en-us/library/ee378677(v=vs.110).aspx

src/Microsoft.Extensions.Caching.Memory/CacheEntry.cs
```diff
@@ -1,47 +1,167 @@
...
         internal CacheEntry(
             object key,
-            object value,
-            DateTimeOffset utcNow,
-            DateTimeOffset? absoluteExpiration,
-            MemoryCacheEntryOptions options,
+            Action<CacheEntry> notifyCacheEntryDisposed,
             Action<CacheEntry> notifyCacheOfExpiration)
         {
+            if (key == null)
```
Tratcher: string.IsNullOrEmpty?

sebastienros: Key is `object`

src/Microsoft.Extensions.Caching.Abstractions/MemoryCacheExtensions.cs
sebastienros: Should we solve it with documentation, on the interface itself, I know it has been an issue for Orchard as even myself was expecting some locking on the factory.

src/Microsoft.Extensions.Caching.Memory/CacheEntry.cs
```diff
@@ -68,14 +205,14 @@ internal void SetExpired(EvictionReason reason)
 
         private bool CheckForExpiredTime(DateTimeOffset now)
         {
-            if (_absoluteExpiration.HasValue && _absoluteExpiration.Value <= now)
+            if (_absoluteExpiration <= now)
```
Tratcher: What does `null <= now` return? false?

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -37,7 +40,7 @@ public MemoryCache(IOptions<MemoryCacheOptions> optionsAccessor)
             var options = optionsAccessor.Value;
             _entries = new Dictionary<object, CacheEntry>();
             _entryLock = new ReaderWriterLockSlim();
-            _entryExpirationNotification = EntryExpired;
```
Tratcher: Why was this removed? I'm told we have to cache these explicitly because otherwise the compiler allocates them every call.

src/Microsoft.Extensions.Caching.Memory/CacheEntry.cs
sebastienros: Yes but I am reverting to prevent any confusion.

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -63,12 +66,8 @@ public int Count
...
         }
 
-        public IEntryLink CreateLinkingScope()
-        {
-            return EntryLinkHelpers.CreateLinkingScope();
-        }
-
-        public object Set(object key, object value, MemoryCacheEntryOptions cacheEntryOptions)
+        /// <inheritdoc />
+        public void Remove(object key)
```
Tratcher: A lot of the diff in this view is just caused by method re-ordering, correct? Can you revert that where possible? E.g. Remove, Dispose, etc.?

sebastienros:
:up: :date:

samples/MemoryCacheSample/Program.cs
```diff
@@ -92,20 +92,20 @@ public static void Main()
             cts.Cancel();
 
             // Expire an entry if the dependent entry expires
-            using (var link = cache.CreateLinkingScope())
+            using (var entry = cache.CreateEntry("key1"))
             {
+                // expire this entry if the entry with key "key2" expires.
+                entry.SetValue("value1");
```
Tratcher: Just show the property if you're not going to use the chaining.

src/Microsoft.Extensions.Caching.Abstractions/MemoryCacheExtensions.cs
```diff
@@ -21,34 +25,78 @@ public static TItem Get<TItem>(this IMemoryCache cache, object key)
...
 
-        public static object Set(this IMemoryCache cache, object key, object value)
+        public static TItem Set<TItem>(this IMemoryCache cache, object key, TItem value)
         {
-            return cache.Set(key, value, new MemoryCacheEntryOptions());
+            return cache.Set(key, value, null);
         }
 
-        public static object Set(this IMemoryCache cache, object key, object value, MemoryCacheEntryOptions options)
+        public static TItem Set<TItem>(this IMemoryCache cache, object key, TItem value, MemoryCacheEntryOptions options)
```
Tratcher: MemoryCacheEntryOptions is now a redundant temporary object. Would it make sense to have a few overloads of Set that just took the most common settings and bypassed creating the MemoryCacheEntryOptions? E.g.:
```
Set<TItem>(this IMemoryCache cache, object key, TItem value, DateTimeOffset absoluteExpiration)
Set<TItem>(this IMemoryCache cache, object key, TItem value, TimeSpan absoluteExpirationRelativeToNow)
Set<TItem>(this IMemoryCache cache, object key, TItem value, IChangeToken expirationToken)
```

src/Microsoft.Extensions.Caching.Memory/CacheEntry.cs
```diff
@@ -1,47 +1,167 @@
...
+
         private static readonly Action<object> ExpirationCallback = ExpirationTokensExpired;
 
         private readonly Action<CacheEntry> _notifyCacheOfExpiration;
 
-        private readonly DateTimeOffset? _absoluteExpiration;
+        private readonly Action<CacheEntry> _notifyCacheEntryDisposed;
 
-        private IList<PostEvictionCallbackRegistration> _postEvictionCallbacks;
+        internal IList<PostEvictionCallbackRegistration> _postEvictionCallbacks;
```
Tratcher: private

sebastienros:
:up: :date:

Added suggested simple extension methods.

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -302,6 +245,60 @@ private void EntryExpired(CacheEntry entry)
             StartScanForExpiredItems();
         }
 
+        private void SetEntry(CacheEntry entry)
```
Tratcher: Move this up with CreateEntry.

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -63,109 +73,52 @@ public int Count
...
+                    _setEntry,
+                    _entryExpirationNotification
+                );
 
-            _entryLock.EnterWriteLock();
+            _entryLock.EnterReadLock();
             try
             {
-                if (_entries.TryGetValue(key, out priorEntry))
+                if (_entries.TryGetValue(entry.Key, out priorEntry))
```
Tratcher: Move into SetEntry. It's not replaced until ICacheEntry.Dispose is called.

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -174,30 +127,19 @@ public bool TryGetValue(object key, out object value)
...
-                            if (entry.Options.ExpirationTokens != null)
-                            {
-                                link.AddExpirationTokens(entry.Options.ExpirationTokens);
-                            }
-                            if (entry.Options.AbsoluteExpiration.HasValue)
-                            {
-                                link.SetAbsoluteExpiration(entry.Options.AbsoluteExpiration.Value);
-                            }
-                        }
+                        // When an entry is returned from an ambient entry,
```
Tratcher: This comment doesn't make sense.
When this entry is retrieved in the scope of creating another entry, that entry needs a copy of these expiration tokens.

sebastienros:
:up: :date:

Moved more code in `SetEntry`, much better now.

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -63,77 +73,61 @@ public int Count
...
+            CacheEntry priorEntry;
 
             _entryLock.EnterWriteLock();
             try
             {
-                if (_entries.TryGetValue(key, out priorEntry))
+                
+                if (_entries.TryGetValue(entry.Key, out priorEntry))
                 {
-                    _entries.Remove(key);
```
Tratcher: Keep this remove. The new entry may already have expired and fail to overwrite, causing SetExpired to be fired again next time.

samples/MemoryCacheSample/Program.cs
```diff
@@ -92,20 +92,20 @@ public static void Main()
...
-                    .AddEntryLink(link)
-                    .RegisterPostEvictionCallback(
+                entry.RegisterPostEvictionCallback(
                     (echoKey, value, reason, substate) =>
                     {
                         Console.WriteLine(echoKey + ": '" + value + "' was evicted due to " + reason);
-                    }));
+                    });
+
+                cache.Set("key2", "value2", new MemoryCacheEntryOptions()
```
Tratcher: There's a new overload for this.

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -314,7 +300,7 @@ private void StartScanForExpiredItems()
                     CancellationToken.None, TaskCreationOptions.DenyChildAttach, TaskScheduler.Default);
             }
         }
-
+
```
Tratcher: nit: whitespace

test/Microsoft.Extensions.Caching.Memory.Tests/CacheServiceExtensionsTests.cs
```diff
@@ -72,7 +72,7 @@ public void AddCaching_DoesNotReplaceUserRegisteredServices()
 
         private class TestMemoryCache : IMemoryCache
         {
-            public IEntryLink CreateLinkingScope()
+            public void Remove(object key)
```
Tratcher: Random reordering?

test/Microsoft.Extensions.Caching.Memory.Tests/MemoryCacheSetAndRemoveTests.cs
```diff
@@ -131,6 +216,39 @@ public void SetOverwritesAndInvokesCallbacks()
         }
 
         [Fact]
+        public void SetOverwritesWithReplacedReason()
+        {
+            //System.Diagnostics.Debugger.Launch();
```
Tratcher: Remove

sebastienros:
:up: :date:

Tratcher:
:shipit:

PR #170, opened 2016-03-28
[Fixes #146] Maximum Key Length on Primary Key

### LINKED ISSUE #146: Maximum Key Length on Primary Key
The ID on the caching table is nvarchar(900); with a maximum key length of 900 bytes won't the ID column have a max size of 1800 byes?

When executing the create table statement, SQL Server does generate a warning.
`Warning! The maximum key length is 900 bytes. The index 'pk_Id' has maximum length of 1800 bytes. For some combination of large values, the insert/update operation will fail.`

kichalla:
@divega @muratg

### DIFF
--- a/samples/SqlServerCacheConcurencySample/project.json
+++ b/samples/SqlServerCacheConcurencySample/project.json
@@ -11,6 +11,10 @@
     "Microsoft.Extensions.PlatformAbstractions": "1.0.0-*"
   },
 
+  "content": [
+    "config.json"
+  ],
+
   "frameworks": {
     "net451": { },
     "netstandardapp1.5": {
--- a/src/dotnet-sql-cache/SqlQueries.cs
+++ b/src/dotnet-sql-cache/SqlQueries.cs
@@ -8,8 +8,10 @@ namespace Microsoft.Extensions.Caching.SqlConfig
     internal class SqlQueries
     {
         private const string CreateTableFormat = "CREATE TABLE {0}(" +
-            // add collation to the key column to make it case-sensitive
-            "Id nvarchar(900) COLLATE SQL_Latin1_General_CP1_CS_AS NOT NULL, " +
+            // Maximum size of primary key column is 900 bytes (898 bytes from key + 2 additional bytes used by the 
+            // Sql Server). In the case where the key is greater than 898 bytes, then it gets truncated.
+            // - Add collation to the key column to make it case-sensitive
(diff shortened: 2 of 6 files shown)


src/dotnet-sql-cache/SqlQueries.cs
```diff
@@ -8,8 +8,10 @@ namespace Microsoft.Extensions.Caching.SqlConfig
     internal class SqlQueries
     {
         private const string CreateTableFormat = "CREATE TABLE {0}(" +
-            // add collation to the key column to make it case-sensitive
-            "Id nvarchar(900) COLLATE SQL_Latin1_General_CP1_CS_AS NOT NULL, " +
+            // - Primary key column cannot exceed the maximum size of 900 bytes. Here each character occupies
+            //   2 bytes.
+            // - Add collation to the key column to make it case-sensitive
+            "Id nvarchar(445) COLLATE SQL_Latin1_General_CP1_CS_AS NOT NULL, " +
```
kichalla: From https://msdn.microsoft.com/en-us/library/ms186939.aspx, the storage size is the (size * 2 bytes) + additional 2 bytes. So this could be just 449 but not sure if we really care it to be that accurate.

samples/SqlServerCacheConcurencySample/project.json
```diff
@@ -11,6 +11,10 @@
     "Microsoft.Extensions.PlatformAbstractions": "1.0.0-*"
   },
```
muratg: Can you make it consistent with the others? (You removed an empty line from the other project.json below.)

src/dotnet-sql-cache/SqlQueries.cs
muratg: If we know the exact number, any concerns just providing that?

divega: I agree with @muratg. These specs are pretty stable and there doesn't seem to be a downside on allowing the maximum key length possible. Maybe add tests that exercise the maximum size and a negative test that exceeds it (do we know what happens when you do?).

kichalla:
:up: :date: Adding tests

src/dotnet-sql-cache/SqlQueries.cs
kichalla: > do we know what happens when you do?
I noticed that if a key greater than 898 bytes is provided, then the value is truncated to 898.

muratg:
:shipit:

PR #175, opened 2016-04-07
Fix package versions

pranavkm:


pranavkm:
cc @davidfowl \ @Eilon

PR #183, opened 2016-04-26
Fix error messages in dotnet-sql-cache tool

kichalla:
@moozzyk 

Sample error messages:

- An error occurred. Invalid Sql server connection string 'a'. Format of the initialization string does not conform to specification starting at index 0.

- An error occurred while trying to create the table and index. The specified schema name "foo" either does not exist or you do not have permission to use it.

### DIFF
--- a/src/Microsoft.Extensions.Caching.SqlConfig.Tools/Program.cs
+++ b/src/Microsoft.Extensions.Caching.SqlConfig.Tools/Program.cs
@@ -23,7 +23,7 @@ namespace Microsoft.Extensions.Caching.SqlConfig.Tools
             loggerFactory.AddConsole();
             _logger = loggerFactory.CreateLogger<Program>();
         }
-        
+
         public static int Main(string[] args)
         {
             return new Program().Run(args);
@@ -67,9 +67,7 @@ namespace Microsoft.Extensions.Caching.SqlConfig.Tools
                         _schemaName = schemaNameArg.Value;
                         _tableName = tableNameArg.Value;
 
-                        CreateTableAndIndexes();
-
-                        return 0;
+                        return CreateTableAndIndexes();
                     });
                 });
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.Extensions.Caching.SqlConfig.Tools/Program.cs
```diff
@@ -124,14 +124,33 @@ private void CreateTableAndIndexes()
...
+                            ex.Message);
                         transaction.Rollback();
+
+                        return 1;
                     }
                 }
             }
         }
+
+        private void ValidateConnectionString(string _connectionString)
```
moozzyk: Is it supposed to take the connection string as a parameter or use the class variable?

src/Microsoft.Extensions.Caching.SqlConfig.Tools/Program.cs
```diff
@@ -124,14 +124,33 @@ private void CreateTableAndIndexes()
                         transaction.Commit();
 
                         _logger.LogInformation("Table and index were created successfully.");
+
+                        return 0;
```
moozzyk: Consider moving this to the end of function

src/Microsoft.Extensions.Caching.SqlConfig.Tools/Program.cs
```diff
@@ -84,13 +82,15 @@ public int Run(string[] args)
             }
             catch (Exception exception)
             {
-                _logger.LogCritical("An error occurred. {Message}", exception.Message);
```
moozzyk: Is there any reason to change it?

src/Microsoft.Extensions.Caching.SqlConfig.Tools/Program.cs
```diff
@@ -84,13 +82,15 @@ public int Run(string[] args)
...
+                _logger.LogCritical("An error occurred. {ErrorMessage}", exception.Message);
                 return 1;
             }
         }
 
-        private void CreateTableAndIndexes()
+        private int CreateTableAndIndexes()
         {
+            ValidateConnectionString(_connectionString);
```
moozzyk: Why do you need it? Wouldn't SqlConnection throw if the connection string is invalid? Note that `connection.Open()` can still throw even if the connection string is syntactically valid.

src/Microsoft.Extensions.Caching.SqlConfig.Tools/Program.cs
```diff
@@ -124,14 +124,33 @@ private void CreateTableAndIndexes()
...
                     }
                 }
             }
         }
+
+        private void ValidateConnectionString(string _connectionString)
+        {
+            try
+            {
+                var connectionStringBuilder = new SqlConnectionStringBuilder(_connectionString);
```
moozzyk: why to assign at all?

src/Microsoft.Extensions.Caching.SqlConfig.Tools/Program.cs
kichalla: Wanted to keep it consistent with line no. 133

src/Microsoft.Extensions.Caching.SqlConfig.Tools/Program.cs
kichalla: > Why do you need it? Wouldn't SqlConnection throw if the connection string is invalid?

True, but I did not want to give false positives for any other exception that SqlConnection might throw when its constructing the object, just wanted to be sure that the validation error corresponds to invalid connection string.

src/Microsoft.Extensions.Caching.SqlConfig.Tools/Program.cs
kichalla: Sure

kichalla:
:up: :date:

src/Microsoft.Extensions.Caching.SqlConfig.Tools/Program.cs
```diff
@@ -103,7 +103,7 @@ private void CreateTableAndIndexes()
                     _logger.LogWarning(
                         $"Table with schema '{_schemaName}' and name '{_tableName}' already exists. " +
                         "Provide a different table name and try again.");
-                    return;
+                    return 1;
                 }
 
                 reader.Dispose();
```
moozzyk: I know it is not related to your change but I just noticed that this will not be called if the query returns non-empty reader

src/Microsoft.Extensions.Caching.SqlConfig.Tools/Program.cs
```diff
@@ -124,14 +124,34 @@ private void CreateTableAndIndexes()
                         transaction.Commit();
 
                         _logger.LogInformation("Table and index were created successfully.");
+
```
moozzyk: nit: remove empty line

kichalla:
:up: :date:

moozzyk:
:ship: :it:

PR #185, opened 2016-05-07
Fixing race condition

### LINKED ISSUE #184: Update value under concurrency fails when a cancellation token is registered.
### Scenario:
1. Set up a memory cache.
2. Add an entry with a cancellation token registered with it.
3. Start a child thread which keep reading the memory cache with the same key.
4. Reset the value of the entry from main thread.
5. Exception `InvalidOperationException` is thrown.

### Issue repo:
https://github.com/troydai/CachingIssues/blob/2a85107b9b572c47d66d2a0303deafecaeaa3166/TestActions/ConcurrencyIssue184.cs#L17

Run the sample follow exception will be thrown from child thread after the value is updated by main thread.

```
Unhandled Exception: System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
   at System.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource)
   at System.Collections.Generic.List`1.Enumerator.MoveNextRare()
   at Microsoft.Extensions.Caching.Memory.CacheEntry.PropageOptions(CacheEntry parent)
   at Microsoft.Extensions.Caching.Memory.MemoryCache.TryGetValue(Object key, Object& result)
   at Microsoft.Extensions.Caching.Memory.CacheExtensions.TryGetValue[TItem](IMemoryCache cache, Object key, TItem& value)
   at Microsoft.Extensions.Caching.Memory.CacheExtensions.Get[TItem](IMemoryCache cache, Object key)
   at ConsoleApplication.Program.<>c__DisplayClass5_0.<Monitor>b__0(Int32 round, DateTime beginning)
   at ConsoleApplication.Program.<>c__DisplayClass4_0.<Polling>b__0()
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
```

sebastienros:
Fixes #184

### DIFF
--- a/src/Microsoft.Extensions.Caching.Memory/CacheEntry.cs
+++ b/src/Microsoft.Extensions.Caching.Memory/CacheEntry.cs
@@ -30,6 +30,7 @@ namespace Microsoft.Extensions.Caching.Memory
         internal DateTimeOffset? _absoluteExpiration;
         internal TimeSpan? _absoluteExpirationRelativeToNow;
         private TimeSpan? _slidingExpiration;
+        private IDisposable _bookmark;
 
         internal readonly object _lock = new object();
 
@@ -57,7 +58,7 @@ namespace Microsoft.Extensions.Caching.Memory
             _notifyCacheEntryDisposed = notifyCacheEntryDisposed;
             _notifyCacheOfExpiration = notifyCacheOfExpiration;
 
-            CacheEntryHelper.EnterScope(this);
+            _bookmark = CacheEntryHelper.EnterScope(this);
         }
 
         /// <summary>
@@ -170,7 +171,7 @@ namespace Microsoft.Extensions.Caching.Memory
             if (!_added)
             {
                 _added = true;
-                CacheEntryHelper.LeaveScope();
+                _bookmark.Dispose();
                 _notifyCacheEntryDisposed(this);
                 PropageOptions(CacheEntryHelper.Current);
             }
--- a/src/Microsoft.Extensions.Caching.Memory/CacheEntry.cs
+++ b/src/Microsoft.Extensions.Caching.Memory/CacheEntry.cs
@@ -52,7 +52,7 @@ namespace Microsoft.Extensions.Caching.Memory
             {
                 throw new ArgumentNullException(nameof(notifyCacheOfExpiration));
             }
-            
+
             Key = key;
             _notifyCacheEntryDisposed = notifyCacheEntryDisposed;
             _notifyCacheOfExpiration = notifyCacheOfExpiration;
@@ -170,9 +170,11 @@ namespace Microsoft.Extensions.Caching.Memory
         internal IList<IDisposable> ExpirationTokenRegistrations { get; set; }
 
         internal DateTimeOffset LastAccessed { get; set; }
-        
+
         public void Dispose()
         {
+            CacheEntryHelper.LeaveScope();
+
             try
             {
                 if (!_added)
@@ -183,7 +185,6 @@ namespace Microsoft.Extensions.Caching.Memory
             }
             finally
             {
-                CacheEntryHelper.LeaveScope();
                 PropageOptions(CacheEntryHelper.Current);
             }
         }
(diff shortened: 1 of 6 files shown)


sebastienros:
Created the PR to ensure all unit tests pass on the CI

src/Microsoft.Extensions.Caching.Memory/CacheEntryHelper.cs
```diff
@@ -51,9 +51,12 @@ internal static CacheEntry Current
             {
                 if (Scopes != null)
                 {
-                    if (Scopes.Count > 0)
```
Tratcher: You shouldn't need any locks in this class, these should all be thread/async local.

src/Microsoft.Extensions.Caching.Memory/CacheEntry.cs
```diff
@@ -334,9 +335,12 @@ internal void PropageOptions(CacheEntry parent)
...
             if (_expirationTokens != null)
             {
-                foreach (var expirationToken in _expirationTokens)
+                lock (_lock)
                 {
-                    parent.AddExpirationToken(expirationToken);
+                    foreach (var expirationToken in _expirationTokens)
+                    {
+                        parent.AddExpirationToken(expirationToken);
+                    }
```
Tratcher: Can you make ExpirationTokenRegistrations private? It's used inside the _lock (elsewhere) and doesn't appear to be accessed externally.

halter73:
@davidfowl So we're removing scopes right?

AsyncLocal/ExecutionContext doesn't play well with spawning new threads and mutating an existing AsyncLocal field independently/concurrently in the subthread(s). Locking fixes the InvalidOperationException, but we could still run into more subtle bugs do to invalid scope chains.

Should we create a new issues to remove scoping from Caching and Logging for RTM?

src/Microsoft.Extensions.Caching.Memory/CacheEntry.cs
```diff
@@ -163,42 +163,30 @@ public IList<PostEvictionCallbackRegistration> PostEvictionCallbacks
...
-                    _added = true;
-                    _notifyCacheEntryDisposed(this);
-                }
-            }
-            finally
+            if (!_added)
             {
+                _added = true;
                 CacheEntryHelper.LeaveScope();
+                _notifyCacheEntryDisposed(this);
```
muratg: Would this throw (asking since it was inside try/catch before)

muratg:
#184 was triaged to 1.0.0, so you'll need to move this to dev branch instead.

halter73:
Definitely something to fix for 1.0.0

@sebastienros pointed out that [serilog has a similar pattern](https://github.com/serilog/serilog/blob/ee68faec21694d27da94555811bb587a64b6c8d0/src/Serilog/Context/LogContext.cs#L59) for tracking scopes that uses an immutable stack.

This works since the ExecutionContext is copy-on-write and so as long as you update the actual AsyncLocal reference each subthread will get it's own copy of the scope stack derived/forked from the parent as intended. We should add tests for this, but in theory this should work exactly as intended.

The performance issues of copying the execution context each time a cache entry is created should probably be measured. At least with logging you only pay if you use the scoping feature, and the console logger does not.

halter73:
Technically, you don't need to create a immutable data structure. You just need not to mutate it after storing it in the AsyncLocal field. Using an immutable collection can guarantee safety though.

test/Microsoft.Extensions.Caching.Memory.Tests/MemoryCacheSetAndRemoveTests.cs
```diff
@@ -368,6 +369,39 @@ public void SetGetAndRemoveWorksWithObjectKeysWhenDifferentReferences()
...
+                {
+                    cache.Set(key, new Guid());
+                }
+            });
+
+            var task2 = Task.Run(() =>
+            {
+                while (!cts2.IsCancellationRequested)
+                {
+                    cache.Get(key);
```
halter73: This verifies that the original issue was fixed (i.e. the IOE is no longer thrown), but it would also be nice to verify that one thread doesn't affect the others view of the `Scopes`.

sebastienros:
Closing to target dev

PR #190, opened 2016-05-17
Work with WeakReferences

### LINKED ISSUE #191: Add special case for WeakReferences
WeakReferences make be GC'd prior to expiry; if there target has been GC'd they should be removed from the cache

benaadams:
Special case and remove WeakReferences if they have been collected before expiry

Resolves #191

### DIFF
--- a/src/Microsoft.Extensions.Caching.Abstractions/EvictionReason.cs
+++ b/src/Microsoft.Extensions.Caching.Abstractions/EvictionReason.cs
@@ -31,5 +31,10 @@ namespace Microsoft.Extensions.Caching.Memory
         /// GC, overflow
         /// </summary>
         Capacity,
+
+        /// <summary>
+        /// GC, weakReference
+        /// </summary>
+        Collected,
     }
 }
--- a/src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
+++ b/src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
@@ -316,6 +316,15 @@ namespace Microsoft.Extensions.Caching.Memory
                     {
                         expiredEntries.Add(entry);
                     }
+                    else
+                    {
+                        var wr = entry.Value as WeakReference;
+                        if (wr != null && !wr.IsAlive)
+                        {
+                            entry.SetExpired(EvictionReason.Collected);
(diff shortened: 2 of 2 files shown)


benaadams:
Don't think it can work with `WeakReference<T>` though :(

sebastienros:
In what context would you cache a `WeakReference` ?

benaadams:
@sebastienros WeakReference to a Response cache item or "nice to have" cache items, basically lightly cached items. e.g. how long do you keep response cache items in memory? Might have thousands of pages all with cache expiry times, yet infrequently accessed

Where you are happy to have them, but don't want them contributing to memory pressure.

Some other types of things http://www.philosophicalgeek.com/2014/09/03/practical-uses-of-weakreference/

sebastienros:
I see, like lower than `Low`, something like `Weak` priority which could use a `WeakReference` internally. Maybe that's how it should be done.

benaadams:
That might be better

benaadams:
That might be better - only issue is it can't return the item if its been evicted in the notification

Tratcher:
What if you made an IChangeToken based on WeakReference, WeakChangeToken? Give it the WeakReference and map HasChanged to IsAlive. Then store the WeakReference in the cache with the WeakChangeToken attached. Whenever MemoryCache scans for expired items it will check HasChanged/IsAlive and remove GC'd items.

omariom:
@benaadams 
> Don't think it can work with  WeakReference<T>  though :(

There is a need for smth like IWeakReference
with ```TryGetValue(out object value);``` and ``` SetValue(object value) ```

Tratcher:
@benaadams how about this instead? https://github.com/aspnet/Caching/commit/08b7f2310cfa7d260470c52ad41d001b04fd8d2a

benaadams:
@Tratcher looks good, though will the expiry scanner drop the entry if `HasChanged` is returning false?

Tratcher:
Oh, my bad, that should be `return !_reference.TryGetTarget(out ignored);`

Tratcher:
Updated: https://github.com/aspnet/Caching/commit/a197fe623325e67f5a9a33f759499bcd4828d235

Tratcher:
Closing this PR in favor of another design.

Issue #169, opened 2016-03-22
IDistributedCache should include StringSet/SetString 


shanselman:
Right now IDistributedCache is a generic interface that's presumably meant to support lots of distributed caches. It's pretty clear, though, that Redis is going to be 80% or more of the usage. Not to mention Azure has a Redis As A Service. 

IDistributedCache has SetAsync which just takes bytes...as it is there's no easy way to add overloads/extensions to get at the underlying Redis implementation and set strings. 

As we are today, you can't set from ASP.NET Core and GET from the CLI. redis-cli supports SET and GET with strings. This makes debugging with the CLI a hassle.

GetString and SetString are atomic enough that they should be added to IDistributedCache. 

/cc @nickcraver @mgravell @DamianEdwards @rustd

Tratcher:
It's worth noting that we don't just store the bytes, we store our own complex object containing the expiration information: https://github.com/aspnet/Caching/blob/dev/src/Microsoft.Extensions.Caching.Redis/RedisCache.cs#L81-L88
So even if we added a string version, I don't know if you'd be able to view it from the CLI. Have you tried to hack it in yet?

mgravell:
Hacky but possibly workable:

    static Task SetAsync(this IDistributedCache cache, string key, string value) {
        return cache.SetAsync(key, Encoding.UTF8.GetBytes(value));
    }

On the CLI front: looking at that script, as long as the caller knows to use `HGETALL {key}` rather than `GET {key}`, that should actually be usable and readable. Unlike our even hackier in-house middleware that is actually a single blob of a protobuf-encoded object that (like you) has optional sliding expiration and *may or may not* have the payload gzip compressed!

As a side note, I so want to C#6-ify that `RedisCache.cs` linked above...

        private const string SetScript = $@"
                redis.call('HMSET', KEYS[1], '{AbsoluteExpirationKey}', ARGV[1], '{SlidingExpirationKey}', ARGV[2], '{DataKey }', ARGV[4])
                if ARGV[3] ~= '{NotPresent }' then
                  redis.call('EXPIRE', KEYS[1], ARGV[3])
                end
                return 1";

;p

shanselman:
Thanks Marc...I guess after "the basics" everyone starts piggybacking
metadata on top or putting it in the serialization format, don't they?

HGETALL works for my basic learning/teaching scenario, thanks!

On Tue, Mar 22, 2016 at 2:51 PM, Marc Gravell <notifications@github.com>
wrote:

[...]

PR #195, opened 2016-05-24
Adding Get/Set String to IDistributedCache

BrennanConroy:
#169 
@JunTaoLuo @Tratcher @Eilon @muratg

### DIFF
--- a/src/Microsoft.Extensions.Caching.Abstractions/DistributedCacheExtensions.cs
+++ b/src/Microsoft.Extensions.Caching.Abstractions/DistributedCacheExtensions.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Text;
 using System.Threading.Tasks;
 
 namespace Microsoft.Extensions.Caching.Distributed
@@ -17,11 +18,16 @@ namespace Microsoft.Extensions.Caching.Distributed
         /// <param name="cache">The cache in which to store the data.</param>
         /// <param name="key">The key to store the data in.</param>
         /// <param name="value">The data to store in the cache.</param>
+        /// <exception cref="System.ArgumentNullException">Thrown when <paramref name="key"/> or <paramref name="value"/> is null.</exception>
         public static void Set(this IDistributedCache cache, string key, byte[] value)
         {
             if (key == null)
             {
                 throw new ArgumentNullException(nameof(key));
+            }
+            if (value == null)
+            {
+                throw new ArgumentNullException(nameof(value));
             }
 
             cache.Set(key, value, new DistributedCacheEntryOptions());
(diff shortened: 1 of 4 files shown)


src/Microsoft.Extensions.Caching.Abstractions/DistributedCacheExtensions.cs
```diff
@@ -43,5 +44,35 @@ public static Task SetAsync(this IDistributedCache cache, string key, byte[] val
 
             return cache.SetAsync(key, value, new DistributedCacheEntryOptions());
         }
+
+        public static void SetString(this IDistributedCache cache, string key, string value)
+        {
+            cache.Set(key, Encoding.UTF8.GetBytes(value));
```
sebastienros: What happens if `value` is `null` ?

src/Microsoft.Extensions.Caching.Abstractions/DistributedCacheExtensions.cs
```diff
@@ -43,5 +44,35 @@ public static Task SetAsync(this IDistributedCache cache, string key, byte[] val
 
             return cache.SetAsync(key, value, new DistributedCacheEntryOptions());
         }
+
+        public static void SetString(this IDistributedCache cache, string key, string value)
```
JunTaoLuo: Doc comments?

src/Microsoft.Extensions.Caching.Abstractions/DistributedCacheExtensions.cs
sebastienros: From the doc it throws, it needs to be handled then.

https://msdn.microsoft.com/en-us/library/ds4kkd55(v=vs.110).aspx

src/Microsoft.Extensions.Caching.Abstractions/DistributedCacheExtensions.cs
Tratcher: You need overloads that take DistributedCacheEntryOptions.

BrennanConroy:
🆙 📅

src/Microsoft.Extensions.Caching.Abstractions/DistributedCacheExtensions.cs
```diff
@@ -43,5 +44,93 @@ public static Task SetAsync(this IDistributedCache cache, string key, byte[] val
...
+        /// Asynchronously sets a string in the specified cache with the specified key.
+        /// </summary>
+        /// <param name="cache">The cache in which to store the data.</param>
+        /// <param name="key">The key to store the data in.</param>
+        /// <param name="value">The data to store in the cache.</param>
+        /// <param name="options">The cache options for the entry.</param>
+        /// <returns>A task that represents the asynchronous set operation.</returns>
+        public static Task SetStringAsync(this IDistributedCache cache, string key, string value, DistributedCacheEntryOptions options)
+        {
+            if (value == null)
```
sebastienros: It should be discussed, but I think we should handle `null` values in a way that a user could really mean "store a null" value, and get a "null" value back. It's a valid value for a string and can have meaning. Even if it means "delete the entry".

And it should not conflict with empty strings either, which are another valid value.

Eilon: Not sure about the "delete the entry" behavior. But I would think we should persist/retrieve `null` strings as a valid value.

Tratcher: There is no ContainsKey, so storying a null entry is indistinguishable from ignoring/removing the entry. We can store string.Empty as byte[0], but we don't have a meaningful way to serialize and store null.

Eilon: We can put some token value to represent `null`. We can store things however we like...

BrennanConroy:
Updated

test/Microsoft.Extensions.Caching.Memory.Tests/MemoryCacheSetAndRemoveTests.cs
```diff
@@ -84,6 +84,16 @@ public void SetAlwaysOverwrites()
         }
 
         [Fact]
+        public void SetNullValueThrows()
```
sebastienros: So we currently can't add a null value to a memory cache, even today?

Eilon: Yeah memory cache should still allow this. Just distributed should block. No?

src/Microsoft.Extensions.Caching.Abstractions/DistributedCacheExtensions.cs
```diff
@@ -34,14 +40,115 @@ public static void Set(this IDistributedCache cache, string key, byte[] value)
...
+        /// <summary>
+        /// Sets a string in the specified cache with the specified key.
+        /// </summary>
+        /// <param name="cache">The cache in which to store the data.</param>
+        /// <param name="key">The key to store the data in.</param>
+        /// <param name="value">The data to store in the cache.</param>
+        /// <exception cref="System.ArgumentNullException">Thrown when <paramref name="key"/> or <paramref name="value"/> is null.</exception>
+        public static void SetString(this IDistributedCache cache, string key, string value)
+        {
+            if (value == null)
```
Tratcher: It's strange to just check the value here. I assume you depend on the called function to check the key? If so, you can do the same for value.

BrennanConroy:
🆙 📅

Tratcher:
:shipit:

PR #196, opened 2016-05-24
Add Caching SqlConfig tool to sample

BrennanConroy:
#186 
@troydai

### DIFF
--- a/samples/SqlServerCacheConcurencySample/project.json
+++ b/samples/SqlServerCacheConcurencySample/project.json
@@ -29,5 +29,10 @@
         "portable-net451+win8"
       ]
     }
+  },
+  "tools": {
+    "Microsoft.Extensions.Caching.SqlConfig.Tools": {
+      "version": "1.0.0-*"
+    }
   }
 }
--- a/samples/SqlServerCacheSample/project.json
+++ b/samples/SqlServerCacheSample/project.json
@@ -28,5 +28,10 @@
         "portable-net451+win8"
       ]
     }
+  },
+  "tools": {
+    "Microsoft.Extensions.Caching.SqlConfig.Tools": {
+      "version": "1.0.0-*"
+    }
   }
 }


Tratcher:
Isn't there an issue with tools that prevents them from being used as project references? e.g. in the same repo. @moozzyk

moozzyk:
Yes. Project to Project references don't work for tools so not sure how you were able to run this. A workaround might be to add a global.json file in the samples folder which will prevent from bringing the tool as project.

BrennanConroy:
Well wouldn't it just download it from myget...

moozzyk:
You can try it (but it did not work in other repos...)

BrennanConroy:
When I ran it that's what it looked like it was doing

BrennanConroy:
Whats up with this?

troydai:
LGTM

moozzyk:
My understanding is that it should not work because of https://github.com/dotnet/cli/issues/1215. I wonder if it might have worked because you had the package cached?

BrennanConroy:
Deleted cache and it still works fine.

Tratcher:
It's still resolving as a package reference, not a project reference, right?

BrennanConroy:
Yeah

moozzyk:
That's why this works. Not sure why it is not resolved as a project reference.

muratg:
@BrennanConroy What's the latest with this?

BrennanConroy:
I can merge this, but its weird because the tool restores from your configured feeds and not as a project reference, there is a CLI issue logged for that

muratg:
@BrennanConroy What's the CLI bug?

BrennanConroy:
https://github.com/dotnet/cli/issues/1215

muratg:
OK, it looks like they backlogged the bug. I'm backlogging this one as well.

Issue #194, opened 2016-05-23
Cannot set SlidingExpiration and AbsoluteExpirationRelativeToNow on RedisCache


camp-007:
When using the RedisCache you can set both a SlidingExpiration and an AbsoluteExpiration.  For example, you can set a sliding expiration of 5 seconds and an absolute expiration of 60 seconds.  If the cached item is accessed every 5 seconds it will remain in cache up until 60 seconds from the original creation.

However, this only works if you are setting the DistributedCacheEntryOptions.AbsoluteExpiration property.  If you set the DistrubutedCacheEntryOption.AbsoluteExpirationRelativeToNow then only the sliding expiration is respected/used.

The meta-data stored in Redis for each cache entry includes only the SlidingExpiration and the AbsoluteExpiration values, but ignores the AbsoluteExpirationRelativeToNow property.

The AbsoluteExpirationRelativeToNow property is very convenient (as the caller often doesn't want to have to worry about calculating the actual expiration time, just specifying how far into the future it should be).  However, it seems that this property should really cause all the same behaviors as the AbsoluteEpiration property.

Thanks for listening!

Tratcher:
Hmm, it should be cheched right here: https://github.com/aspnet/Caching/blob/67aef94080bc0302ece7a2b42f5a97ed68efdfa1/src/Microsoft.Extensions.Caching.Redis/RedisCache.cs#L326

camp-007:
Correct, it is checked there in GetExpirationInSeconds(), however, that is only used to initially set the TTL/expiry on the redis key.  The way the sliding expiration works, is that each "Get" will use the GetAndRefresh() method to update the key TTL/expiry.  This is done by getting the SlidingExpiration and AbsoluteExpiration out of the metadata and passing them into the refresh method (shown here):

https://github.com/aspnet/Caching/blob/67aef94080bc0302ece7a2b42f5a97ed68efdfa1/src/Microsoft.Extensions.Caching.Redis/RedisCache.cs#L178

But the problem, I think, is that when the meta data is stored, it only accounts for the AbsoluteExpiration property (and it ignores the AbsoluteExpirationRelativeToNow property).  This can be seen here:

https://github.com/aspnet/Caching/blob/67aef94080bc0302ece7a2b42f5a97ed68efdfa1/src/Microsoft.Extensions.Caching.Redis/RedisCache.cs#L84

[...]

PR #197, opened 2016-05-25
Fix RelativeExpiration with SlidingExpiration

BrennanConroy:
#194 
@Tratcher

### DIFF
--- a/src/Microsoft.Extensions.Caching.Redis/RedisCache.cs
+++ b/src/Microsoft.Extensions.Caching.Redis/RedisCache.cs
@@ -78,12 +78,14 @@ namespace Microsoft.Extensions.Caching.Redis
 
             var creationTime = DateTimeOffset.UtcNow;
 
+            var absoluteExpiration = GetAbsoluteExpiration(creationTime, options);
+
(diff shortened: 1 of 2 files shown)


src/Microsoft.Extensions.Caching.Redis/RedisCache.cs
```diff
@@ -78,10 +78,16 @@ public void Set(string key, byte[] value, DistributedCacheEntryOptions options)
 
             var creationTime = DateTimeOffset.UtcNow;
 
+            var absoluteExpiration = options.AbsoluteExpiration;
+            if (options.AbsoluteExpirationRelativeToNow.HasValue)
+            {
+                absoluteExpiration = creationTime + options.AbsoluteExpirationRelativeToNow;
```
Tratcher: Should this absoluteExpiration be passed into GetExpirationInSeconds to reduce redundant logic there?

BrennanConroy: Yeah, I was thinking something like that

BrennanConroy:
🆙 📅

Tratcher:
:shipit:

PR #198, opened 2016-05-26
Fixing potential cache key collisions between IDistributedCache and IMemoryCache

### LINKED ISSUE #176: MemoryDistributedCache should prefix keys to avoid conflicts with IMemoryCache
I am having a scenario where I first look in a local `IMemoryCache` for a hit, and if not found, go to a distribted `IDistributedCache`. If the key is found in the distributed cache, I store is locally for a short time in the `IMemoryCache`.
This works perfectly when `IDistributedCache` is e.g. a Redis cache, but when the implementation is `MemoryDistributedCache`, then the keyspace for the two caches is the same. This results in the first lookup in `IMemoryCache` to return the entry expected to be returned in `IDistributedCache`.

Would it be possible to e.g. prefix all keys in `MemoryDistributedCache` with something like "distributed" so that the two caches behaves like they are separate.

sebastienros:
Fixes #176


Tratcher:
:shipit:

PR #199, opened 2016-05-27
Fix OSX build on Travis.

cesarbs:


Tratcher:
What in Caching depends on OpenSsl? SQL?

cesarbs:
It's not Caching itself. We need to perform those tests before installing dotnet in OSX: https://www.microsoft.com/net/core#macosx

Tratcher:
Ah, so every repo needs this?

cesarbs:
Yes.

Issue #203, opened 2016-06-07
System.AppDomainUnloadedException when running UnitTests with RC2


divega:
_From @Mertsch on June 3, 2016 12:20_

After the upgrade from RC1 to RC2 some of my Unit Tests which test against the DB fail with the following exception.
The behavior is reproducible, BUT only happens in a certain combination of tests. So I can't give a direct example as I am unable to find out which combination of my 90 tests causes this.
Fact is each tests runs fine on its own, all ran fine together with RC1, all tests run fine under R# Test runner, TFS (VSTS) / Test in VS show the same behavior.

The line where the test fails is a `DBContext` accessor, but many other `DBContext` operations have been successful before the specific one.

Someone else seems to have the same issue
http://stackoverflow.com/questions/37378698/ef-core-rc2-appdomainunloadedexception
I am not using any In-Memory provider, only pure SQL Server

Each Unit Test builds up a completely new DI context, so that nothing is shared between Tests in any (static) way

```
Error Message:
   Test method xxx.ServiceTests.CleanupSuggestedTest threw exception: 
System.AppDomainUnloadedException: Attempted to access an unloaded AppDomain.
Stack Trace:
    at System.Runtime.Remoting.ObjectHandle.Unwrap()
   at Microsoft.Extensions.Caching.Memory.CacheEntryHelper.get_Scopes()
   at Microsoft.Extensions.Caching.Memory.CacheEntryHelper.EnterScope(CacheEntry entry)
   at Microsoft.Extensions.Caching.Memory.MemoryCache.CreateEntry(Object key)
   at Microsoft.Extensions.Caching.Memory.CacheExtensions.Set[TItem](IMemoryCache cache, Object key, TItem value)
   at Microsoft.EntityFrameworkCore.Query.Internal.CompiledQueryCache.GetOrAddAsyncQuery[TResult](Object cacheKey, Func`1 compiler)
   at Microsoft.EntityFrameworkCore.Query.Internal.QueryCompiler.ExecuteAsync[TResult](Expression query, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.Query.Internal.EntityQueryProvider.ExecuteAsync[TResult](Expression expression, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.EntityFrameworkQueryableExtensions.ExecuteAsync[TSource,TResult](MethodInfo operatorMethodInfo, IQueryable`1 source, CancellationToken cancellationToken)
[...]

PR #213, opened 2016-07-14
Fix exception when changing AppDomains

BrennanConroy:
#203


davidfowl:
We need to look at logging as well and the `HttpContextAccessor`

test/Microsoft.Extensions.Caching.Memory.Tests/MemoryCacheSetAndRemoveTests.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.Extensions.Caching.Memory
 {
     public class MemoryCacheSetAndRemoveTests
     {
-        private IMemoryCache CreateCache()
+        static private IMemoryCache CreateCache()
```
Tratcher: private static.

Tratcher:
wow :shipit:

PR #215, opened 2016-07-18
[Design] [Fixes #209] SQLCache Set should allow a null timeout

### LINKED ISSUE #209: SQLCache Set should allow a null timeout
I have implemented sql cache:

```
using Microsoft.Extensions.Caching.Distributed;
using Microsoft.Extensions.Caching.SqlServer;

.
.
.

public void ConfigureServices(IServiceCollection services)
 {
      services.AddSingleton<IDistributedCache>(serviceProvider =>
      new SqlServerCache(
            new CacheOptions( new SqlServerCacheOptions()
            {
                  ConnectionString = @"Data Source=localhost;Initial Catalog=CacheDB;Integrated Security=False;User Id=sa;Password=pippo",
                  SchemaName = "dbo",
                  TableName = "testCache"
            })
      ));
 }

public void Configure(IApplicationBuilder app,  IDistributedCache cache)
{
      var serverStartTimeString = DateTime.Now.ToString();
      byte[] val = Encoding.UTF8.GetBytes(serverStartTimeString);
      cache.Set("lastServerStartTime", val);

      app.Run(async (context) =>
      {
            context.Response.ContentType = "text/html";
            await context.Response.WriteAsync("Hello World!");
      });
}


```
but when I execute the code, I receive the error "Either absolute or sliding expiration needs to be provided.".
This because at line 33 of [Microsoft.Extensions.Caching.Abstractions/DistributedCacheExtensions.cs](https://github.com/aspnet/Caching/blob/dev/src/Microsoft.Extensions.Caching.Abstractions/DistributedCacheExtensions.cs) the  DistributedCacheEntryOptions parameter is a new value, with no setting  for AbsoluteExpirationRelativeToNow or SlidingExpiration property.

If i use 
```

cache.Set("lastServerStartTime", val, new DistributedCacheEntryOptions()
{
      AbsoluteExpirationRelativeToNow = new TimeSpan(1,0,0)
});

```
All works fine.

Need to modify Microsoft.Extensions.Caching.Abstractions/DistributedCacheExtensions.cs  to pass default value forAbsoluteExpirationRelativeToNow or SlidingExpiration ??

kichalla:
@Tratcher

### DIFF
--- a/src/Microsoft.Extensions.Caching.SqlServer/SqlServerCache.cs
+++ b/src/Microsoft.Extensions.Caching.SqlServer/SqlServerCache.cs
@@ -22,6 +22,7 @@ namespace Microsoft.Extensions.Caching.SqlServer
         private readonly TimeSpan _expiredItemsDeletionInterval;
         private DateTimeOffset _lastExpirationScan;
         private readonly Action _deleteExpiredCachedItemsDelegate;
+        private readonly TimeSpan _defaultSlidingExpiration;
 
         public SqlServerCache(IOptions<SqlServerCacheOptions> options)
         {
@@ -49,11 +50,19 @@ namespace Microsoft.Extensions.Caching.SqlServer
                     $"{nameof(SqlServerCacheOptions.ExpiredItemsDeletionInterval)} cannot be less the minimum " +
                     $"value of {MinimumExpiredItemsDeletionInterval.TotalMinutes} minutes.");
             }
+            if (cacheOptions.DefaultSlidingExpiration <= TimeSpan.Zero)
+            {
+                throw new ArgumentOutOfRangeException(
+                    nameof(cacheOptions.DefaultSlidingExpiration),
+                    cacheOptions.DefaultSlidingExpiration,
+                    "The sliding expiration value must be positive.");
+            }
 
             _systemClock = cacheOptions.SystemClock ?? new SystemClock();
             _expiredItemsDeletionInterval =
                 cacheOptions.ExpiredItemsDeletionInterval ?? DefaultExpiredItemsDeletionInterval;
             _deleteExpiredCachedItemsDelegate = DeleteExpiredCacheItems;
+            _defaultSlidingExpiration = cacheOptions.DefaultSlidingExpiration;
 
             // SqlClient library on Mono doesn't have support for DateTimeOffset and also
             // it doesn't have support for apis like GetFieldValue, GetFieldValueAsync etc.
@@ -169,6 +178,8 @@ namespace Microsoft.Extensions.Caching.SqlServer
                 throw new ArgumentNullException(nameof(options));
             }
 
+            GetOptions(ref options);
+
             _dbOperations.SetCacheItem(key, value, options);
 
             ScanForExpiredItemsIfRequired();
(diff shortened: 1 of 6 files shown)


src/Microsoft.Extensions.Caching.SqlServer/SqlServerCacheOptions.cs
```diff
@@ -37,6 +46,11 @@ public class SqlServerCacheOptions : IOptions<SqlServerCacheOptions>
         /// </summary>
         public string TableName { get; set; }
 
+        /// <summary>
+        /// The default options for a new cache entry. It has SlidingExpiration set for 20 minutes by default.
+        /// </summary>
+        public DistributedCacheEntryOptions DefaultCacheEntryOptions { get; set; }
```
Tratcher: Don't use this type here, it's not future proof, just add `DefaultSlidingExpiration`.

Tratcher: E.g. we may add per-entry options that don't make sense to set globally.

Tratcher:
Waiting for user feedback on #209.

src/Microsoft.Extensions.Caching.SqlServer/SqlServerCacheOptions.cs
kichalla: Ah good point

iscifoni:
For me is good.

src/Microsoft.Extensions.Caching.SqlServer/SqlServerCache.cs
```diff
@@ -164,10 +173,7 @@ public void Set(string key, byte[] value, DistributedCacheEntryOptions options)
                 throw new ArgumentNullException(nameof(value));
             }
 
-            if (options == null)
```
Tratcher: Keep the null check, we want a consistent contract across the IDistributedCache APIs and all the existing extensions already pass in an instance. https://github.com/aspnet/Caching/blob/dev/src/Microsoft.Extensions.Caching.Abstractions/DistributedCacheExtensions.cs#L33

Tratcher: Looks like we need to add the null checks to redis cache or it will null ref in GetAbsoluteExpiration

Tratcher: The same for MemoryDistributedCache.Set

kichalla:
:up: :date:

src/Microsoft.Extensions.Caching.SqlServer/SqlServerCache.cs
```diff
@@ -164,11 +173,13 @@ public void Set(string key, byte[] value, DistributedCacheEntryOptions options)
...
+            if(options == null)
```
Tratcher: space?

test/Microsoft.Extensions.Caching.SqlServer.Tests/SqlServerCacheWithDatabaseTest.cs
```diff
@@ -10,14 +10,13 @@
...
 using Microsoft.Extensions.Internal;
-using Microsoft.Extensions.Options;
 using Xunit;
 
 namespace Microsoft.Extensions.Caching.SqlServer
 {
     // This requires SQL Server database to be setup
     // public
-    class SqlServerCacheWithDatabaseTest
+    public class SqlServerCacheWithDatabaseTest
```
Tratcher: were these intentionally disabled because they require SQL?

Tratcher:
:shipit:

PR #217, opened 2016-08-12
Sample for caching with weak references

Tratcher:
#191 @benaadams @sebastienros


Tratcher:
@kichalla @muratg

muratg:
:shipit:

sebastienros:
Should it be part of a separate package or namespace and also include some other useful tokens like a `string` based one to invalidate by `string`? c.f. https://github.com/aspnet/Caching/issues/219

PR #229, opened 2016-09-08
Bump tools version

natemcmaster:
Merges to dev. This will help us avoid possible confusion between versions built from `dev` and `rel/tools-1.0.0-preview3` (cref #227 ) 

cc @muratg @pranavkm


muratg:
:shipit:

PR #232, opened 2016-09-19
Move dotnet-sql-cache to aspnet/DotNetTools repo

natemcmaster:
Moving day! dotnet-sql-cache is moving to aspnet/DotNetTools

Destination: https://github.com/aspnet/dotnet-watch/pull/164

cc @Eilon @muratg


Eilon:
cc @kichalla

PR #238, opened 2016-10-11
Ensure Redis and SqlServer cache operations silently fail

### LINKED ISSUE #193: More reliable cache operations
As part of discussions while working on https://github.com/aspnet/Session/issues/99, there were proposals for adding logic in caching to make database operations more reliable. For example, sql database operations may fail intermittently. In this case it may be a good idea to add retry logic and allow users to specify retry policies. This probably needs some more design work.

cc @mikeharder @Tratcher @davidfowl

natemcmaster:
Resolve #193 

Per our design discussion, this adds top-level catches for all exceptions with optional logging for suppressed exceptions.

NB: introduces dependency on logging abstractions.

cc @JunTaoLuo @Tratcher @pakrym

### DIFF
--- a/src/Microsoft.Extensions.Caching.Redis/RedisCache.cs
+++ b/src/Microsoft.Extensions.Caching.Redis/RedisCache.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Caching.Distributed;
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using StackExchange.Redis;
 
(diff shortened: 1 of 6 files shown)


src/Microsoft.Extensions.Caching.Redis/RedisCache.cs
```diff
@@ -380,6 +426,14 @@ private void Refresh(string key, DateTimeOffset? absExpr, TimeSpan? sldExpr)
             return absoluteExpiration;
         }
 
+        private void LogSuppressedException(Exception ex)
+        {
+            if (_logger != null)
+            {
+                _logger.LogDebug("An exception during cache operation was suppressed: {message}", ex.ToString());
```
Tratcher: you don't need to insert the message, you can pass down the entire exception object.

src/Microsoft.Extensions.Caching.Redis/RedisCache.cs
```diff
@@ -380,6 +426,14 @@ private void Refresh(string key, DateTimeOffset? absExpr, TimeSpan? sldExpr)
             return absoluteExpiration;
         }
 
+        private void LogSuppressedException(Exception ex)
+        {
+            if (_logger != null)
+            {
+                _logger.ExceptionSuppressed(ex);
```
pakrym: optional nit: you can do `_logger?.ExceptionSuppressed(ex);` in catch blocks

pakrym:
:shipit:

JunTaoLuo:
Would be nice to have a few tests to make sure exceptions are caught and the correct messages are logged.

natemcmaster:
Reverted in 7abe1ccc0aab2f43066e82f3ba5bff2c40ea41d8

Issue #242, opened 2016-10-20
Improve performance of MemoryCache by using ConcurrentDictionary rather than ReaderWriterLockSlim


mikeharder:
It may be possible to significantly improve the performance of MemoryCache by using `ConcurrentDictionary` rather than `Dictionary` protected by a `ReaderWriterLockSlim`.  Under high load, there is contention acquiring the read lock from `ReaderWriterLockSlim`.

I created a quick prototype which shows significantly improved performance, though it may not be functionally correct in all scenarios.

| Scenario | MemoryCache Implementation | RPS | CPU |
| --- | --- | --- | --- |
| Plaintext | N/A | 1,167,369 | 100% |
| MemoryCachePlaintext | ReaderWriterLockSlim | 342,129 | 60% |
| MemoryCachePlaintext | ConcurrentDictionary | 802,076 | 100% |

The remaining work is to improve and test the ConcurrentDictionary implementation to ensure it's correct under high load, has no race conditions, etc.

Prototype: https://github.com/aspnet/Caching/tree/mikeharder/concurrent-dictionary

Dissimilis:
Please thoroughly test ConcurrentDictionary vs lock'ed Dictionary performance and don't make this change blindly. Performance may depend a lot on specific workload (amount of get vs set, etc.), even simple Monitor may be faster than ReaderWriterLockSlim if most operations are writes.

JunTaoLuo:
@Dissimilis thanks for the feedback but is there any particular workload you are concerned about? A few scenarios we have benchmarked include 1) concurrent gets with a few set to update the entry and 2) concurrent sets and removes and have seen major improvements in both. See the referenced PR for benchmark results. The move toward `ConcurrentDictionary` also takes advantage of the more fine-grained locking and will significantly improve performance when concurrently accessing different entries in the cache compared to the previous implementation which uses a single `ReaderWriterLockSlim` across the entire dictionary.

sebastienros:
Can we update the table with the new numbers?

JunTaoLuo:
Scenario | previous CPU | previous RPS | CPU | RPS
--- | --- | --- | --- | ---
benchmarks: memorycacheplaintext | 60 | 342,129 | 100 | 886,544
benchmarks: memorycacheplaintextsetremove | - | - | 100 | 689,634

PR #246, opened 2016-10-31
Concurrent dictionary implementation for MemoryCache

JunTaoLuo:
#242

Performance results:

Scenario | Before | After
--- | --- | ---
benchmarks: memorycacheplaintext | 280249.98 | 886543.73
benchmarks: memorycacheplaintextsetremove | 123328.57 | 689634.29

### DIFF
--- a/src/Microsoft.Extensions.Caching.Memory/CacheEntry.cs
+++ b/src/Microsoft.Extensions.Caching.Memory/CacheEntry.cs
@@ -1,12 +1,6 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-#if NETSTANDARD1_3
-#else
-using System.Runtime.Remoting;
-using System.Runtime.Remoting.Messaging;
-#endif
-
 using System;
 using System.Collections.Generic;
 using System.Threading;
@@ -22,7 +16,6 @@ namespace Microsoft.Extensions.Caching.Memory
         private readonly Action<CacheEntry> _notifyCacheOfExpiration;
         private readonly Action<CacheEntry> _notifyCacheEntryDisposed;
         private IList<IDisposable> _expirationTokenRegistrations;
-        private EvictionReason _evictionReason;
         private IList<PostEvictionCallbackRegistration> _postEvictionCallbacks;
         private bool _isExpired;
 
@@ -166,6 +159,8 @@ namespace Microsoft.Extensions.Caching.Memory
 
         internal DateTimeOffset LastAccessed { get; set; }
 
+        internal EvictionReason EvictionReason { get; private set; }
+
         public void Dispose()
         {
             if (!_added)
@@ -184,11 +179,11 @@ namespace Microsoft.Extensions.Caching.Memory
 
         internal void SetExpired(EvictionReason reason)
         {
-            _isExpired = true;
-            if (_evictionReason == EvictionReason.None)
+            if (EvictionReason == EvictionReason.None)
             {
-                _evictionReason = reason;
+                EvictionReason = reason;
             }
+            _isExpired = true;
             DetachTokens();
         }
 
@@ -302,7 +297,7 @@ namespace Microsoft.Extensions.Caching.Memory
 
                 try
                 {
-                    registration.EvictionCallback?.Invoke(entry.Key, entry.Value, entry._evictionReason, registration.State);
+                    registration.EvictionCallback?.Invoke(entry.Key, entry.Value, entry.EvictionReason, registration.State);
                 }
                 catch (Exception)
                 {
--- a/src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
+++ b/src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
@@ -72,6 +72,8 @@ namespace Microsoft.Extensions.Caching.Memory
             get { return _entries.Count; }
         }
 
+        private ICollection<KeyValuePair<object, CacheEntry>> EntriesCollection => _entries;
+
         /// <inheritdoc />
         public ICacheEntry CreateEntry(object key)
         {
@@ -86,6 +88,12 @@ namespace Microsoft.Extensions.Caching.Memory
 
         private void SetEntry(CacheEntry entry)
         {
+            if (_disposed)
+            {
+                // No-op instead of throwing since this is called during CacheEntry.Dispose
+                return;
+            }
+
             var utcNow = _clock.UtcNow;
 
             DateTimeOffset? absoluteExpiration = null;
@@ -112,31 +120,57 @@ namespace Microsoft.Extensions.Caching.Memory
             // Initialize the last access timestamp at the time the entry is added
             entry.LastAccessed = utcNow;
 
-            var added = false;
             CacheEntry priorEntry;
-
-            if (_entries.TryRemove(entry.Key, out priorEntry))
+            if (_entries.TryGetValue(entry.Key, out priorEntry))
             {
                 priorEntry.SetExpired(EvictionReason.Replaced);
             }
 
             if (!entry.CheckExpired(utcNow))
             {
-                if (_entries.TryAdd(entry.Key, entry))
+                var entryAdded = false;
+
+                if (priorEntry == null)
+                {
+                    // Try to add the new entry if no previous entries exist.
+                    entryAdded = _entries.TryAdd(entry.Key, entry);
+                }
+                else
+                {
+                    // Try to update with the new entry if a previous entries exist.
+                    entryAdded = _entries.TryUpdate(entry.Key, entry, priorEntry);
+
+                    if (!entryAdded)
+                    {
+                        // The update will fail if the previous entry was removed after retrival.
+                        // Adding the new entry will succeed only if no entry has been added since.
+                        // This guarantees removing an old entry does not prevent adding a new entry.
+                        entryAdded = _entries.TryAdd(entry.Key, entry);
+                    }
+                }
+
+                if (entryAdded)
                 {
                     entry.AttachTokens();
-                    added = true;
+                }
+                else
+                {
+                    entry.SetExpired(EvictionReason.Replaced);
+                    entry.InvokeEvictionCallbacks();
+                }
+
+                if (priorEntry != null)
+                {
+                    priorEntry.InvokeEvictionCallbacks();
                 }
             }
-
-            if (priorEntry != null)
-            {
-                priorEntry.InvokeEvictionCallbacks();
-            }
-
-            if (!added)
+            else
             {
                 entry.InvokeEvictionCallbacks();
+                if (priorEntry != null)
+                {
+                    RemoveEntry(priorEntry);
+                }
             }
 
             StartScanForExpiredItems();
@@ -150,18 +184,21 @@ namespace Microsoft.Extensions.Caching.Memory
                 throw new ArgumentNullException(nameof(key));
             }
 
-            var utcNow = _clock.UtcNow;
-            result = null;
-            bool found = false;
-            CacheEntry expiredEntry = null;
             CheckDisposed();
+
+            result = null;
+            var utcNow = _clock.UtcNow;
+            var found = false;
+
             CacheEntry entry;
             if (_entries.TryGetValue(key, out entry))
             {
                 // Check if expired due to expiration tokens, timers, etc. and if so, remove it.
-                if (entry.CheckExpired(utcNow))
+                // Allow a stale Replaced value to be returned due to a race with SetExpired during SetEntry.
+                if (entry.CheckExpired(utcNow) && entry.EvictionReason != EvictionReason.Replaced)
                 {
-                    expiredEntry = entry;
+                    // TODO: For efficiency queue this up for batch removal
+                    RemoveEntry(entry);
                 }
                 else
                 {
@@ -173,12 +210,6 @@ namespace Microsoft.Extensions.Caching.Memory
                     // that entry needs a copy of these expiration tokens.
                     entry.PropagateOptions(CacheEntryHelper.Current);
                 }
-            }
-
-            if (expiredEntry != null)
-            {
-                // TODO: For efficiency queue this up for batch removal
-                RemoveEntry(expiredEntry);
             }
 
             StartScanForExpiredItems();
@@ -196,15 +227,10 @@ namespace Microsoft.Extensions.Caching.Memory
 
             CheckDisposed();
             CacheEntry entry;
-            if (_entries.TryGetValue(key, out entry))
+            if (_entries.TryRemove(key, out entry))
             {
                 entry.SetExpired(EvictionReason.Removed);
-            }
-
-            if (entry != null)
-            {
-                // TODO: For efficiency consider processing these removals in batches.
-                RemoveEntry(entry);
+                entry.InvokeEvictionCallbacks();
             }
 
             StartScanForExpiredItems();
@@ -212,30 +238,7 @@ namespace Microsoft.Extensions.Caching.Memory
 
         private void RemoveEntry(CacheEntry entry)
         {
-            // Only remove it if someone hasn't modified it since our lookup
-            CacheEntry currentEntry;
-            if (_entries.TryGetValue(entry.Key, out currentEntry)
-                && object.ReferenceEquals(currentEntry, entry))
-            {
-                _entries.TryRemove(entry.Key, out currentEntry);
-            }
-            entry.InvokeEvictionCallbacks();
-        }
-
-        private void RemoveEntries(List<CacheEntry> entries)
-        {
-            foreach (var entry in entries)
-            {
-                // Only remove it if someone hasn't modified it since our lookup
-                CacheEntry currentEntry;
-                if (_entries.TryGetValue(entry.Key, out currentEntry)
-                    && object.ReferenceEquals(currentEntry, entry))
-                {
-                    _entries.TryRemove(entry.Key, out currentEntry);
-                }
-            }
-
-            foreach (var entry in entries)
+            if (EntriesCollection.Remove(new KeyValuePair<object, CacheEntry>(entry.Key, entry)))
             {
                 entry.InvokeEvictionCallbacks();
             }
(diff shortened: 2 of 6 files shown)


src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -162,40 +157,26 @@ public bool TryGetValue(object key, out object result)
...
+                    entry.LastAccessed = utcNow;
+                    result = entry.Value;
 
-            if (expiredEntry != null)
-            {
-                // TODO: For efficiency queue this up for batch removal
-                RemoveEntry(expiredEntry);
+                    // When this entry is retrieved in the scope of creating another entry,
+                    // that entry needs a copy of these expiration tokens.
+                    entry.PropagateOptions(CacheEntryHelper.Current);
```
JunTaoLuo: This looks suspicious for concurrency.

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -87,6 +86,12 @@ public ICacheEntry CreateEntry(object key)
 
         private void SetEntry(CacheEntry entry)
         {
+            if (_disposed)
+            {
+                // No-op instead of throwing since this is called during Dispose
```
Tratcher: CacheEntry.Dispose?

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -113,37 +118,19 @@ private void SetEntry(CacheEntry entry)
...
-                    _entries[entry.Key] = entry;
-                    entry.AttachTokens();
-                    added = true;
-                }
-            }
-            finally
-            {
-                _entryLock.ExitWriteLock();
+                priorEntry.SetExpired(EvictionReason.Replaced);
+                priorEntry.InvokeEvictionCallbacks();
```
Tratcher: behavior change: we don't want to invoke the callbacks until the new entry has been added.

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -113,37 +118,19 @@ private void SetEntry(CacheEntry entry)
...
             }
 
-            if (priorEntry != null)
+            if (!entry.CheckExpired(utcNow))
             {
-                priorEntry.InvokeEvictionCallbacks();
+                _entries[entry.Key] = entry;
```
Tratcher: not enough thread safety. you have a race between sets where an item may get replaced without having its callbacks invoked.

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -213,73 +188,32 @@ public void Remove(object key)
...
-                // Only remove it if someone hasn't modified it since our lookup
-                CacheEntry currentEntry;
-                if (_entries.TryGetValue(entry.Key, out currentEntry)
-                    && object.ReferenceEquals(currentEntry, entry))
-                {
-                    _entries.Remove(entry.Key);
-                }
-            }
-            finally
+            if (((ICollection<KeyValuePair<object, CacheEntry>>)_entries).Remove(new KeyValuePair<object, CacheEntry>(entry.Key, entry)))
```
Tratcher: my eyes!!!
Store the `ICollection<KeyValuePair<object, CacheEntry>>` as another member field so you don't have to do this cast every time.

Tratcher: did you check that this does reference equality on the entries?

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -213,73 +188,32 @@ public void Remove(object key)
...
             {
-                _entryLock.ExitWriteLock();
+                entry.InvokeEvictionCallbacks();
             }
-            entry.InvokeEvictionCallbacks();
         }
 
         private void RemoveEntries(List<CacheEntry> entries)
```
Tratcher: This method has been simplified enough that you may be able to inline it. Is it called in more than one place?

JunTaoLuo: Not after I removed a call. I think it's only called in one place now.

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
JunTaoLuo: I thought it wasn't important to keep the order but I suppose it makes sense to avoid behaviour changes.

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
JunTaoLuo: Will use TryAdd instead.

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -72,6 +73,19 @@ public int Count
             get { return _entries.Count; }
         }
 
+        private ICollection<KeyValuePair<object, CacheEntry>> EntriesCollection
```
Tratcher: No need to do this lazily, you can do it in the constructor.

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -122,19 +136,22 @@ private void SetEntry(CacheEntry entry)
             if (_entries.TryRemove(entry.Key, out priorEntry))
             {
                 priorEntry.SetExpired(EvictionReason.Replaced);
-                priorEntry.InvokeEvictionCallbacks();
             }
 
-            if (!entry.CheckExpired(utcNow))
+            if (!entry.CheckExpired(utcNow) && _entries.TryAdd(entry.Key, entry))
```
Tratcher: if !TryAdd you need to set the eviction reason to Replaced

test/Microsoft.Extensions.Caching.Memory.Tests/CacheEntryScopeExpirationTests.cs
```diff
@@ -376,8 +376,8 @@ public void LinkContextsCanNest()
 
             Assert.Equal(1, ((CacheEntry)entry1)._expirationTokens.Count());
             Assert.Null(((CacheEntry)entry1)._absoluteExpiration);
-            Assert.Equal(1, ((CacheEntry)entry1)._expirationTokens.Count());
-            Assert.Null(((CacheEntry)entry1)._absoluteExpiration);
+            Assert.Equal(1, ((CacheEntry)entry)._expirationTokens.Count());
```
JunTaoLuo: I'm assuming this was a typo?

JunTaoLuo:
The performance results have been updated and I've been stress testing the memory cache for concurrency issues. So far I haven't ran into any problems. Looking to do more verification but in the mean time I think the change is ready for review.

cc @sebastienros @mikeharder @Tratcher @davidfowl

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -113,41 +120,34 @@ private void SetEntry(CacheEntry entry)
             // Initialize the last access timestamp at the time the entry is added
             entry.LastAccessed = utcNow;
 
-            var added = false;
             CacheEntry priorEntry;
-
-            _entryLock.EnterWriteLock();
-            try
+            if (_entries.TryRemove(entry.Key, out priorEntry))
```
JunTaoLuo: Instead of TryRemove and TryAdd, we can use TryGet and TryUpdate. That should avoid a concurrent read reading a value of null.

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -213,70 +201,18 @@ public void Remove(object key)
...
-                    }
-                }
-            }
-            finally
-            {
-                _entryLock.ExitWriteLock();
-            }
-
-            foreach (var entry in entries)
+            if (_entriesCollection.Remove(new KeyValuePair<object, CacheEntry>(entry.Key, entry)))
```
pakrym: Why do it via ICollection instead of `TryRemove`?

JunTaoLuo: Because we want to make sure no updates to the entry has been made before we remove it.

pakrym: I don't think `ICollection<KeyValuePair<TKey, TValue>>.Remove` checks this: https://github.com/dotnet/corefx/blob/master/src/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentDictionary.cs#L1334

pakrym: My bad, it does check.

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -43,8 +44,8 @@ public MemoryCache(IOptions<MemoryCacheOptions> optionsAccessor)
             }
 
             var options = optionsAccessor.Value;
-            _entries = new Dictionary<object, CacheEntry>();
-            _entryLock = new ReaderWriterLockSlim();
+            _entries = new ConcurrentDictionary<object, CacheEntry>();
```
mikeharder: Why have two fields pointing to the same object, instead of just casting when needed?

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -87,6 +88,12 @@ public ICacheEntry CreateEntry(object key)
 
         private void SetEntry(CacheEntry entry)
         {
+            if (_disposed)
+            {
+                // No-op instead of throwing since this is called during CacheEntry.Dispose
+                return;
+            }
+
```
mikeharder: Is this related to the `ConcurrentDictionary` change, or was it a bug in the previous implementation as well?

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
pakrym: Would `TryUpdate` work if there is no existing entry?

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
Tratcher: https://github.com/aspnet/Caching/pull/246#discussion-diff-85819069R202

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
JunTaoLuo: Need to check implementation to see if tryupdate can update null. Otherwise maybe use tryadd instead of tryupdate if previous value is null.

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
mikeharder: Is this for perf or readability?  If it's just the latter, personally I find two fields pointing to the same object more confusing than doing the cast when needed.  Do we do this anywhere else in our codebase?

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
pakrym: From what I understand it can update null, but it cannot update when there is no entry at all. (https://github.com/dotnet/corefx/blob/master/src/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentDictionary.cs#L577)

mikeharder: I think the most important thing is to add a test which verifies that the current `ConcurrentDictionary` behavior is incorrect, and the new behavior is correct.  Since it's a race condition, you can't have a unit test which guarantees failure, but I think one thread reading while another is writing should hit the race quickly enought.

pakrym: I depends on what we think is correct. Before MemoryCache was blocking if there was write operation on the way so you were guaranteed to get latest value. In current state of this PR you can get `null` in some cases, in change proposed by @JunTaoLuo you would get old value instead null. I don't think we can achieve exactly same behavior that `ReaderWriterLock` had so we need to decide how much is enough.

mikeharder: I agree that readers shouldn't get `null` if a writer is updating.  If a reader and writer are racing, of course it's fine for the reader to get the old or new value.  What is the observable difference between `ReaderWriterLock` and the new change to `ConcurrentDictionary`?

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
JunTaoLuo: It was aesthetics since the cast is rather verbose. There's no functional reason why we can't just cast as needed.

mikeharder: In that case, I would use a private get-only property like 'EntriesAsCollection' which encapsulates the cast to improve readability, rather than adding another field.

test/Microsoft.Extensions.Caching.Memory.Tests/MemoryCacheSetAndRemoveTests.cs
```diff
@@ -410,37 +410,49 @@ public void SetGetAndRemoveWorksWithObjectKeysWhenDifferentReferences()
         }
 
         [Fact]
-        public void GetAndSet_AreThreadSafe()
+        public void GetAndSet_AreThreadSafe_AndUpdatesNeverLeavesNullValues()
```
JunTaoLuo: Updating this test to handle the null scenario. Could split it into a separate test but I don't think that's necessary.

test/Microsoft.Extensions.Caching.Memory.Tests/MemoryCacheSetAndRemoveTests.cs
```diff
@@ -410,37 +410,49 @@ public void SetGetAndRemoveWorksWithObjectKeysWhenDifferentReferences()
...
             {
-                await Task.Delay(TimeSpan.FromSeconds(1));
-                cts2.Cancel();
+                await Task.Delay(TimeSpan.FromSeconds(10));
             });
 
+            Task.WaitAny(task1, task2, task3);
+
+            Assert.False(readValueIsNull);
+            Assert.Equal(TaskStatus.Running, task1.Status);
```
JunTaoLuo: three redundant asserts, but they don't hurt.

src/Microsoft.Extensions.Caching.Memory/CacheEntry.cs
```diff
@@ -182,13 +176,19 @@ internal bool CheckExpired(DateTimeOffset now)
             return _isExpired || CheckForExpiredTime(now) || CheckForExpiredTokens();
         }
 
+        internal bool CheckExpired(DateTimeOffset now, bool allowReplacedEntries)
+        {
+            var entryExpired = allowReplacedEntries ? _isExpired && _evictionReason != EvictionReason.Replaced : _isExpired;
```
JunTaoLuo: Race here.

Tratcher: `if (entry.CheckExpired(time) && entry.Reason != Reason.Removed)`

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -113,41 +120,38 @@ private void SetEntry(CacheEntry entry)
...
+                    entry.InvokeEvictionCallbacks();
                 }
             }
-            finally
+            else
             {
-                _entryLock.ExitWriteLock();
+                if (priorEntry != null)
+                {
+                    RemoveEntry(priorEntry);
```
Tratcher: note this is a partial duplicate with lines 150-153

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -159,43 +163,31 @@ public bool TryGetValue(object key, out object result)
...
+            var utcNow = _clock.UtcNow;
+            var found = false;
+
+            CacheEntry entry;
+            if (_entries.TryGetValue(key, out entry))
             {
-                CacheEntry entry;
-                if (_entries.TryGetValue(key, out entry))
+                // Check if expired due to expiration tokens, timers, etc. and if so, remove it.
+                if (entry.CheckExpired(utcNow) && entry.EvictionReason != EvictionReason.Replaced)
```
Tratcher: "Allow a stale Replaced value to be returned due to a race with SetEntry."

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
JunTaoLuo: Yes but it's safe to call `InvokeEvictionCallbacks()` more than once. It should just no-op. Alternatively, `EntriesCollection.Remove(...)` can be used here instead

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -113,41 +120,38 @@ private void SetEntry(CacheEntry entry)
...
-            try
+            if (_entries.TryGetValue(entry.Key, out priorEntry))
             {
+                priorEntry.SetExpired(EvictionReason.Replaced);
+            }
 
-                if (_entries.TryGetValue(entry.Key, out priorEntry))
+            if (!entry.CheckExpired(utcNow))
+            {
+                if (priorEntry == null ? _entries.TryAdd(entry.Key, entry) : _entries.TryUpdate(entry.Key, entry, priorEntry))
```
JunTaoLuo: Potential race with removal on expiry.

JunTaoLuo:
There was a race between `SetEntry` and `RemoveEntry` (from expiration, capacity, ...) that prevented a new entry from being added if the old one is removed. I'm taking the approach of an additional `TryAdd` after a failed `TryUpdate` since it produced the highest performance (651K RPS) versus two alternatives: 1) Retry the logic in `SetEntry` in a loop (624K RPS) and 2) write lock `SetEntry` and `RemoveEntry` (478K RPS). Still need to make sure no additional races are introduced with the added call to `TryAdd`.

JunTaoLuo:
ping @Tratcher @pakrym

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -113,39 +120,57 @@ private void SetEntry(CacheEntry entry)
...
+                    {
+                        // The update will fail if the previous entry was removed after retrival.
+                        // Adding the new entry will succeed only if no entry has been added since.
+                        // This guarantees removing an old entry does not prevent adding a new entry.
+                        entryAdded = _entries.TryAdd(entry.Key, entry);
+                    }
                 }
 
-                if (!entry.CheckExpired(utcNow))
+                if (entryAdded)
```
pakrym: Can a race happen between `entry.AttachTokens` and `priorEntry.InvokeEvictionCallbacks` called by other SetEntry operation?

JunTaoLuo: Spoke offline, There's no race here. We don't need to attach expiration tokens if the entry has already been removed and eviction callbacks are invoked.

Issue #221, opened 2016-08-22
Remove GC Compaction trigger


muratg:
@KKhurin @davidfowl 

Looks like we may have room for improvement here.

muratg:
@KKhurin Are you able to help us with this?

KKhurin:
Let's have a discussion.

DamianEdwards:
I just hit what I think is a down side of the current implementation in that it appears we purge the cache on every Gen2 collection, which in my case was occurring during app start as I was forcibly priming my cache.

JunTaoLuo:
Currently compaction is triggered by a gen2 collection by the GC. This is not the best trigger since it may occur frequently and does not correlate to the actual memory usage of the cache.

The proposal is to remove the current trigger and instead rely on the user or external triggers to decide when to compact the cache. This will require us to expose the Compact on the interface. Note that there's currently no way of accurately determining the memory usage by the cache so it's difficult to figure out if or how much compaction should occur.

Eilon:
@JunTaoLuo can you update the bug title to reflect the decisions that were made (e.g. change Foo to X because Y).

JunTaoLuo:
After the latest discussions, we decided to remove the memory compaction trigger based on Gen2 GC. To compact, downcast to `MemoryCache` and call `Compact` when needed.

JunTaoLuo:
Will file follow up issue for revisiting the memory cache design.

Eilon:
@JunTaoLuo is this change announcement-worthy?

JunTaoLuo:
Yup that's why there is a breaking changes tag. Will get to it.

muratg:
@JunTaoLuo Do you want to keep the bug open in that case?

JunTaoLuo:
I will open a different issue in the Discussion milestone for the discussion to the announcement.

code-mogwai:
Just checking the workaround for version 1.0.0 is to set CompactOnMemoryPressure to false for the MemoryCache instance?

(FYI, what we think we are experiencing as a result of this design bug is poor caching and only 30% memory utilisation on a server with 16GB RAM. Another server with 32GB RAM and the same application is also at 30% memory utilisation.)

JunTaoLuo:
@code-mogwai yes, if this is affecting you, you should set `CompactOnMemoryPressure` to false. If you need to compact, you can downcast to `MemoryCache` and call `Compact` directly.

PR #252, opened 2016-12-10
Add to MemoryCacheOptions the CustomCompactOnMemoryPressureDelegate.

Wallsmedia:
### Add to MemoryCacheOptions the CustomCompactOnMemoryPressureDelegate.

The reason of adding  the custom compact delegate:
1. I have tried the feature MemoryCacheOptions.CompactOnMemoryPressure = true and found that it is a good concept to trigger of revising the memory cache size when the GC collects generation 2 .
2. However, the compact implementation is to reduce the cache every time by 10%, it is a weird solution. By the short time it will remove everything from the cache. In the real working .net core rest web service, the cached elements have disappeared before it will be reused.  So, I don't have a cache.

3. I propose to introduce in the MemoryCacheOptions the CustomCompactOnMemoryPressureDelegate as 
`
		public Action<MemoryCache> CustomCompactOnMemoryPressureDelegate { get; set; }
`	
a. If you don't set this delegate the old Compact(0.10) will be invoked.
b. If setup you setup the delegate, it will be invoked instead.

	
For example when the Memory Cache is reached the dedicated number 110000 only in this case the compact method will be called.
	
Example of the custom delegate:

```
 private void CustomDelegateCompact(MemoryCache cache)
	        {
	            if(cache.Count > 110000)
	            {
	                cache.Compact(0.1);
	            }
	        }

```


Tratcher:
We actually have concerns about this trigger mechanic. See https://github.com/aspnet/Caching/issues/221. I don't think we can evaluate your PR until we've resolved our concerns with the trigger.

Wallsmedia:
I came through the thread #221.
My solution has been addressed to solve the similar case. 
I had the similar issues in my web rest service.

Good in triggering:
1. The idea of  triggering mechanic  had been borrowing from [Gc.ReRegisterForFinalize](https://msdn.microsoft.com/EN-US/library/h4s82506(v=VS.110,d=hv.2).aspx)  , see example. 
2.  The calling of garbage collector for generation #2 and collecting  elements from generation #2 it is the best moment to review the size of the memory cache. 

Issues in implementation of triggering:
1. The calling of collect generation #2 can be happened more often in case an ASP Web service. Chances are very high to have non cached objects which are  survived to generation #2 and released and collected.
2. The compact  has  a hard coded compress ration  of 10%
3. The compact triggers after every generation #2 collection.
3. The triggering mechanism of compacting can be called very often. 
   By other words, the ratio may be "1 added to cache" : "> 1 call of compact "

Solution to the issues:
( Not sure if you don't want support backward compatibility by adding features  that are not destroy current contract. )

1. Add ability to define the custom delegate that will be called by triggering mechanism . In this custom delegate you can define a required logic when to call compact and how match to remove.
  
If you check the test cases in my pull request, you will see the solution to the #221 issue.

davidfowl:
Best to close this issue. We're making some design changes that will allow this and we're moving away from this mechanic in general

JunTaoLuo:
We decided to remove the GC based eviction trigger in #221. We will revisit the design of `MemoryCache` in #289.

Wallsmedia:
It is very nice and fair decision. I will be happy to take part in the pre-re-view  coming changes. I will subscribe on further news/changes.

Issue #253, opened 2016-12-12
Redis: Connection Leak if Connect or ConnectAsync called from multiple threads simultaneously


JonCole:
We have a customer who, under load testing, noticed connections to Redis spiking significantly.  After digging into the dump of the process, we could see that many ConnectionMultiplexer objects exist in the heap.  We also noticed that there were many threads calling Connect at the same time.  

If multiple threads make it past the if (_connection == null) check, then there will be a connection leak for each additional thread that makes it through that check.

Tratcher:
Relevant code: https://github.com/aspnet/Caching/blob/1.0.0/src/Microsoft.Extensions.Caching.Redis/RedisCache.cs#L146-L162

@JonCole what version of the packages are they using? Do we need to backport this fix to 1.1.x or 1.0.x?

/cc: @muratg

JonCole:
@Tratcher - looks like they are on 1.1.0.21115.

muratg:
Putting this in 1.1.1 milestone so that it gets tracked with the rest of 1.1.1 candidates.  We should also look into whether we want this in 1.0.4 as well.

JunTaoLuo:
Fixed in https://github.com/aspnet/Caching/commit/3522082828cd0c411f297d50db289518f7844dee. Will need to port this to 1.1.1 branch when available for release.

runxc1:
Would this only happen on startup?  I am noticing about once a day or so on a site with 1 million hits a day that Redis will run out of connections but just for 5-10 minutes

Eilon:
This patch bug is **approved**. Please use the normal code review process w/ a PR and make sure the fix is in the correct branch, then close the bug and mark it as done.

JunTaoLuo:
Patch merged. Leaving the issue open since version numbers will need to be updated.

runxc1:
What schedule is this project on for a a new release push to Nuget?

Eilon:
@runxc1 we're hoping to get these patches out in February.

MikaelPorttila:
*Bump* Any release date yet?

runxc1:
@MikaelPorttila I am following a couple rather large issues in DotNetCore and it seems that all projects are going to ship at the same time as everyone has quoted "February" as when they are going to fix the issues I am watching.  I had thought part of the advantage of doing it in the open and disconnecting everything was to be able to release out of bound but it seems that everything is still on a schedule.

Eilon:
[...]

PR #254, opened 2016-12-13
Lock on creating redis connections

JunTaoLuo:
#253

### DIFF
--- a/src/Microsoft.Extensions.Caching.Redis/RedisCache.cs
+++ b/src/Microsoft.Extensions.Caching.Redis/RedisCache.cs
@@ -29,13 +29,13 @@ namespace Microsoft.Extensions.Caching.Redis
         private const string DataKey = "data";
         private const long NotPresent = -1;
 
-        private ConnectionMultiplexer _connection;
+        private volatile ConnectionMultiplexer _connection;
         private IDatabase _cache;
 
         private readonly RedisCacheOptions _options;
         private readonly string _instance;
 
-        private readonly SemaphoreSlim _connectionLock = new SemaphoreSlim(1, 1);
+        private readonly SemaphoreSlim _connectionLock = new SemaphoreSlim(initialCount: 1, maxCount: 1);
 
         public RedisCache(IOptions<RedisCacheOptions> optionsAccessor)
         {
@@ -158,6 +158,11 @@ namespace Microsoft.Extensions.Caching.Redis
 
         private void Connect()
         {
+            if (_connection != null)
+            {
+                return;
+            }
+
             _connectionLock.Wait();
             try
             {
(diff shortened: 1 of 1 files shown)


Tratcher:
@JonCole

src/Microsoft.Extensions.Caching.Redis/RedisCache.cs
```diff
@@ -34,6 +35,8 @@ public class RedisCache : IDistributedCache, IDisposable
         private readonly RedisCacheOptions _options;
         private readonly string _instance;
 
+        private readonly SemaphoreSlim _connectionLock = new SemaphoreSlim(1, 1);
```
Tratcher: named parameters please

JonCole:
@JunTaoLuo,  

I would recommend that you structure the fix something like the following psuedo-code to avoid lock contention when the connection has already been created.

```
if (connection is null)
{
    //avoid taking the lock in the cases where the connection 
   //has already been created.
    TakeLock();  
    {
        if(connection is null)
        {
              CreateConnection()
        }
    }     
}
```

JunTaoLuo:
@JonCole Yup that works, I was just talking to @sebastienros about the exact same thing.

JunTaoLuo:
@muratg we need to decide whether this goes into 1.0.4 as well.

cc @mikeharder

muratg:
@JunTaoLuo Please file a separate bug for 1.0.4. (and link it from #253).

Also, we should keep #253 open after this is checked in to dev branch.  For 1.1.1, we'll need to port it to some rel/* branch as 1.1.1 approaches.

mikeharder:
The code looks correct.  Personally, I would extract the initialization logic into a helper class, named something like `AsyncLazy<T>`.  Similar to `Lazy<T>` in the framework , but supporting both sync and async initialization.  Then the code in this class would just be the following:

```
private AsyncLazy<IDatabase> _cache = new AsyncLazy<IDatabase>(() => /* sync init code */, () => /* async init code */);

var cache = _cache.Value();
var cache2 = await _cache.AsyncValue();
```

You shouldn't even need to keep an explicit reference to the ConnectionMultiplexer, since the IDatabase reference will prevent it from being GC.  This change would also eliminate the need to call `Connect()` before other operations.

src/Microsoft.Extensions.Caching.Redis/RedisCache.cs
```diff
@@ -28,12 +29,14 @@ public class RedisCache : IDistributedCache, IDisposable
         private const string DataKey = "data";
         private const long NotPresent = -1;
 
-        private ConnectionMultiplexer _connection;
+        private volatile ConnectionMultiplexer _connection;
```
davidfowl: Why is this volatile?

PR #256, opened 2016-12-13
Interlock.exchange to handle locking around the connection creation

roshantarudkar:
There was no locking around the connection creation hence added Interlock.exchange to handle this.


Tratcher:
This change is already in progress. See https://github.com/aspnet/Caching/pull/254

roshantarudkar:
@Tratcher When shall we expect this changes in final version?

Tratcher:
We're considering this for 1.0.4/1.1.1
https://github.com/aspnet/Home/wiki/Roadmap

PR #262, opened 2017-01-11
Concurrent updates to parent requires synchronization

### LINKED ISSUE #247: AddExpirationToken extension is not thread safe
I've noticed that `CacheEntryScopeExpirationTests.LinkContexts_AreThreadSafe` test can occasionally fail when run repeatedly at https://github.com/aspnet/Caching/blob/dev/test/Microsoft.Extensions.Caching.Memory.Tests/CacheEntryScopeExpirationTests.cs#L475. 

I will usually see one or two failures when repeating this test 50 times. I have reproduced this on Win10 x64 and Win10 x86.

cc @Tratcher

JunTaoLuo:
Fixes #247

### DIFF
--- a/src/Microsoft.Extensions.Caching.Memory/CacheEntry.cs
+++ b/src/Microsoft.Extensions.Caching.Memory/CacheEntry.cs
@@ -320,9 +320,12 @@ namespace Microsoft.Extensions.Caching.Memory
             {
                 lock (_lock)
                 {
-                    foreach (var expirationToken in _expirationTokens)
+                    lock (parent._lock)
                     {
-                        parent.AddExpirationToken(expirationToken);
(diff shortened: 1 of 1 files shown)


src/Microsoft.Extensions.Caching.Memory/CacheEntry.cs
```diff
@@ -322,7 +322,10 @@ internal void PropagateOptions(CacheEntry parent)
                 {
                     foreach (var expirationToken in _expirationTokens)
                     {
-                        parent.AddExpirationToken(expirationToken);
+                        lock (parent._lock)
```
Tratcher: why not lock once outside of the foreach?

JunTaoLuo:
:up::date:

PR #265, opened 2017-01-20
Remove Gen2 GC based trigger for compacting memory cache #221

JunTaoLuo:
#221 

Will need approval from @Eilon and @DamianEdwards due to the breaking API change.


Eilon:
Do we have any gauge as to whether anyone out there has implemented this API and would be broken?

src/Microsoft.Extensions.Caching.Abstractions/IMemoryCache.cs
```diff
@@ -30,5 +30,11 @@ public interface IMemoryCache : IDisposable
         /// </summary>
         /// <param name="key">An object identifying the entry.</param>
         void Remove(object key);
+
+        /// <summary>
+        /// Attempt to remove the given percentage of the total entries in the cache.
+        /// </summary>
+        /// <param name="percentage">The percentage of cache entries to remove.</param>
+        void Compact(double percentage);
```
Eilon: Can you update the doc comment in `MemoryCache` itself to be a proper doc comment? Right now there are some odd `///` comments there but not an actual doc comment. https://github.com/aspnet/Caching/blob/cd57aea9755dc8cbf0dbfa1cfa35babfd78f49d2/src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs#L275-L281

Eilon: Also, the implementation in `MemoryCache` doesn't do any argument validation, so that needs to be fixed as well.

Tratcher: should be return the number of elements removed? or a bool indicating if any elements were removed?

JunTaoLuo: Since the argument is in percentage, it would probably make more sense to return the percentage of elements removed?

Tratcher: That doesn't seem very informative. Also, percentages are problematic if you remove 0/0.

JunTaoLuo: Also, since we are exposing Compact, should we expose Count on the interface as well? It's not a good way to gauge cache size but I suspect users will downcast and rely on it to determine how much to compact like in https://github.com/aspnet/Caching/pull/252.

Tratcher: The point of using percentages is that they don't need to know the size. Otherwise Compact would take a specific amount.

Or we shift it so Compact takes a specific number of elements to purge, returns the actual number removed, and expose Count.

sebastienros: > The point of using percentages is that they don't need to know the size

I would say that knowing what percentage you need is also improbable. I would just provide a strategy instead. Like FIFO or based on usage. This strategy would then have access to all the calls to track what it needs (count, memory, usage frequency of each entry, ...). This strategy would be responsible to call Compact internally based on its own logic. And if I want a strategy that is dumb and just removes a percentage of random items then I could.

JunTaoLuo: Shifting to expose Count and using Compact to use specific numbers might be a good change.

@sebastienros we still have policies it's just not configurable or extensible right now. The Compact method removes based on 1) whether the entry is expired 2) CacheItemPriority 3) LRU. There was discussion on incorporating memory footprint into the triggering mechanism but the problem is that there's currently no accurate way of determining memory usage. As such it was determined to expose Compact on the interface and allow users or external triggers to compact the memory cache directly.

Eilon: I'm pretty certain we don't want a Count property on the cache - it's utterly meaningless.

JunTaoLuo: Then let's keep using percentages for both the amount to remove and amount removed. In the 0/0 case, just return 0. Returning the number of entry removed is rather strange considering a percentage was passed in. Returning a bool is also very uninformative.

Tratcher: @Eilon except we've punted the decide-when-to-compact duty to the developer, and the number of elements in the cache is one inputs you'd use to decide when and how much to purge.

Tratcher: @JunTaoLuo put this on hold until we can sync in person Monday.

JunTaoLuo: Will probably solicit feedback from @davidfowl and @DamianEdwards as well if possible.

Eilon: I don't think anyone can ever reason about how much to clear based on the number of items in the cache. Remember, the cache is generally a shared resource. Someone (including our own codes) might have stashed a thousand small objects there. Or 1 really ginormous one.

But yes, let's discuss.

JunTaoLuo:
We're going with a different design. I will update the issue and open a new PR.

Wallsmedia:
@JunTaoLuo Could have a look a my proposal, 
Add to MemoryCacheOptions the CustomCompactOnMemoryPressureDelegate. #252
It may contribute to design.
I have added example how it woks.
To be clear, I have been using this approach/solution in my web services.

davidfowl:
We'll use it as an input but we're basically separating the concept of memory pressure detection out of the cache and adding a first class way to provide compaction logic via an interface. There will also be a way to trigger a compaction via an API exposed from the cache.

Similar so what you delegate can do but less specific.

Issue #257, opened 2016-12-16
Concurrent connection on redis cache leads to connection leak


JunTaoLuo:
Need to port fix in https://github.com/aspnet/Caching/commit/3522082828cd0c411f297d50db289518f7844dee to 1.0.4 branch when available.

muratg:
(Setting milestone to 1.0.2 to reflect the actual package version to be shipped.)

Eilon:
This patch bug is **approved**. Please use the normal code review process w/ a PR and make sure the fix is in the correct branch, then close the bug and mark it as done.

Eilon:
@muratg please assign this bug.

JunTaoLuo:
Patch merged. Leaving the issue open since version numbers will need to be updated.

JunTaoLuo:
Version numbers updated

PR #267, opened 2017-01-21
Lock on creating redis connections

JunTaoLuo:
Patch PR for #257

### DIFF
--- a/src/Microsoft.Extensions.Caching.Redis/RedisCache.cs
+++ b/src/Microsoft.Extensions.Caching.Redis/RedisCache.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Caching.Distributed;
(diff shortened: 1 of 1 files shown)


src/Microsoft.Extensions.Caching.Redis/RedisCache.cs
```diff
@@ -28,12 +29,14 @@ public class RedisCache : IDistributedCache, IDisposable
         private const string DataKey = "data";
         private const long NotPresent = -1;
 
-        private ConnectionMultiplexer _connection;
+        private volatile ConnectionMultiplexer _connection;
```
benaadams: Drop `volatile` and `SemaphoreSlim` and just do a double checked lock?
```csharp
if (_connection == null)
{
    lock(_connectionLock)
    {
        if (_connection == null)
        {
            // ...
        }
    }
}
```

JunTaoLuo: Connect**Async**?

benaadams: Ahhh... 👍

PR #280, opened 2017-03-10
MemoryCache eviction restructure

JunTaoLuo:
#221 Need to review API changes and then continue with additional todos

TODO:
- [ ] tests
- [ ] doc comments
- [ ] perf
- [ ] samples?


src/Microsoft.Extensions.Caching.Memory/IMemoryCacheEvictionStrategy.cs
```diff
@@ -0,0 +1,14 @@
...
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Extensions.Caching.Memory
+{
+    public interface IMemoryCacheEvictionStrategy
+    {
+        // TODO: doc comments
+        IEnumerable<CacheEntry> GetEntriesToEvict(IEnumerable<CacheEntry> entries, DateTimeOffset now);
```
davidfowl: Should this be `IList<CacheEntry>`? This structure worries me a little because it requires the caller to do a LINQ query (or some equivalent) by default. We should add a micro benchmark to see what the performance overhead of this is.

JunTaoLuo: Yea although this is the most generic collection, it may be a little too limited in terms of operations that are available on it.

src/Microsoft.Extensions.Caching.Memory/DefaultMemoryCacheEvictionStrategy.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.Extensions.Caching.Memory
+{
+    public class DefaultMemoryCacheEvictionStrategy : IMemoryCacheEvictionStrategy
```
Tratcher: Avoid the word Default in type names

JunTaoLuo:
Any actionable changes to the API before I continue working on this @davidfowl @Tratcher ?

src/Microsoft.Extensions.Caching.Memory/CacheEntry.cs
```diff
@@ -9,7 +9,7 @@
 
 namespace Microsoft.Extensions.Caching.Memory
 {
-    internal class CacheEntry : ICacheEntry
+    public class CacheEntry : ICacheEntry
```
Tratcher: CacheEntry is the wrong view to expose to the eviction algorithm. This view is the one used for creating the entry so Value is settable, all the expirations are settable, AbsoluteExpirationRelativeToNow is relative to entry creation rather than the expiration scan, etc..

There should be a different interface/view exposed for creation vs eviction, even if the same object implements both internally.

src/Microsoft.Extensions.Caching.Memory/IMemoryCacheEvictionStrategy.cs
Tratcher: Rather than returning a list, maybe just call a method / set  a property on a given entry to mark it for eviction. `Evict()`

Tratcher: Actually, SetExpired pretty much already does that.

src/Microsoft.Extensions.Caching.Abstractions/CacheEntryExtensions.cs
```diff
@@ -148,7 +136,6 @@ public static ICacheEntry SetOptions(this ICacheEntry entry, MemoryCacheEntryOpt
             entry.AbsoluteExpiration = options.AbsoluteExpiration;
             entry.AbsoluteExpirationRelativeToNow = options.AbsoluteExpirationRelativeToNow;
             entry.SlidingExpiration = options.SlidingExpiration;
-            entry.Priority = options.Priority;
```
Tratcher: Removing Priority is unfortunate but I understand. Do we have another way to add custom metadata for the eviction algorithm without embedding it in the value itself? What if ICacheEntry had a transparent `object EvictionMetadata` property?

JunTaoLuo: Not sure about this since EvictionMetadata may not make sense when all components share the same memory cache in DI. Would each component then have to know what to set the EvictionMetadata to? Also, what is the overhead of embedding the information in the value itself over having this field (would need to test this in a microbenchmark).

src/Microsoft.Extensions.Caching.Memory/CacheEntry.cs
JunTaoLuo: I'll give this a try

src/Microsoft.Extensions.Caching.Memory/IMemoryCacheEvictionStrategy.cs
davidfowl: I like that.

JunTaoLuo: That would work.

src/Microsoft.Extensions.Caching.Abstractions/CacheEntryExtensions.cs
Tratcher: All components share an eviction strategy anyways, and if the metadata were embedded in the value you'd have the same problem of everyone having to know about it.

src/Microsoft.Extensions.Caching.Memory/MemoryCacheEvictionTrigger.cs
```diff
@@ -0,0 +1,97 @@
...
+                {
+                    _timerIsRunning = false;
+                    _timer.Change(Timeout.Infinite, Timeout.Infinite);
+                }
+            }
+        }
+
+        private void TimerLoop(object state)
+        {
+            if (EvictionCallback())
```
JunTaoLuo: Need to have logic to prevent this from being re-entrant.

src/Microsoft.Extensions.Caching.Abstractions/CacheEntryExtensions.cs
Tratcher: Sample?

Tratcher:
Think about the error handling for this new extensibility, especially on the timer thread.

src/Microsoft.Extensions.Caching.Memory/MemoryCacheOptions.cs
```diff
@@ -11,9 +11,9 @@ public class MemoryCacheOptions : IOptions<MemoryCacheOptions>
     {
         public ISystemClock Clock { get; set; }
 
-        public bool CompactOnMemoryPressure { get; set; } = true;
+        public IMemoryCacheEvictionStrategy EvictionStrategy { get; set; }
```
Tratcher: Why are these options rather than DI injected?

src/Microsoft.Extensions.Caching.Memory/Interfaces/IMemoryCacheEvictionTrigger.cs
```diff
@@ -0,0 +1,16 @@
...
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Extensions.Caching.Memory
+{
+    public interface IMemoryCacheEvictionTrigger : IDisposable
+    {
+        // TODO: doc comments
+        Func<bool> EvictionCallback { get; set; }
```
Tratcher: `void Initialize(Func<bool> evictionCallback);` or make it a parameter to Resume

src/Microsoft.Extensions.Caching.Memory/CacheEntry.cs
```diff
@@ -172,12 +170,12 @@ public void Dispose()
...
         }
 
-        internal bool CheckExpired(DateTimeOffset now)
+        public bool CheckExpired(DateTimeOffset now)
         {
             return _isExpired || CheckForExpiredTime(now) || CheckForExpiredTokens();
         }
 
-        internal void SetExpired(EvictionReason reason)
+        public void SetExpired(EvictionReason reason)
```
Tratcher: It's a little odd for this to be public. The only EvictionReason that would make sense for the eviction algorithm to set would be Capacity. Expired is the next best reason, but CheckExpired should set that. Maybe make a new method that's just Evict and always uses Capacity as the reason?

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -16,9 +13,9 @@ namespace Microsoft.Extensions.Caching.Memory
     /// An implementation of <see cref="IMemoryCache"/> using a dictionary to
     /// store its entries.
     /// </summary>
-    public class MemoryCache : IMemoryCache
+    public partial class MemoryCache : IMemoryCache
```
Tratcher: partial?

JunTaoLuo:
Closing this PR, we decided to remove the Gen2 GC trigger and allow users to downcast to use Compact on MemoryCache instead.

PR #283, opened 2017-03-22
Disable API Check in project that breaks the tool

dougbu:
- hits a `FileNotFoundException` because System.Data.SqlClient package lacks runtime assemblies for .NET Standard 1.2


dougbu:
Will go in before fix for aspnet/KoreBuild#143

dougbu:
See aspnet/BuildTools#205 for more context

dougbu:
afa938c

PR #286, opened 2017-03-24
Remove old test

BrennanConroy:
@pranavkm 
There are similar cases of this in a few repos, I saw you didn't remove the test in Mvc when doing the net46 change. Haven't checked other repos yet.


pranavkm:
Do we want to not want to keep the tests?

BrennanConroy:
This was only an issue when dealing with `CallContext`

pranavkm:
:shipit:

PR #302, opened 2017-05-01
Use the bundled NETStandard.Library package in netstandard targeting libraries

pranavkm:


natemcmaster:
Aha, this condition:
```
And '$(_TargetFrameworkVersionWithoutV)' == '$(BundledNETStandardTargetFrameworkVersion)'
```

https://github.com/dotnet/sdk/blob/release/2.0.0/src/Tasks/Microsoft.NET.Build.Tasks/build/Microsoft.NET.Sdk.DefaultItems.targets#L53-L59

nguerrera:
@natemcmaster That condition is there so that you don't pick up an unnecessary pre-release dependency while targeting netstandard1.x. Why do you need a newer netstandard package without targeting netstandard2.0?

natemcmaster:
> Why do you need a newer netstandard package without targeting netstandard2.0?

I'm not sure, actually. @pranavkm or @davidfowl? Why are we using netstandard.library 2.0.0-preview1 when targeting netstandard1.x?

pranavkm:
@DamianEdwards \ @Eilon 's edict. We're going to change our desktop targeting projects to target `netstandard2.0` in preview2.

Eilon:
I believe it's been discussed many times that in general you need to:

1. Target the lowest `netstandardX.Y` that you can: This gets your the broadest reach (obviously)
2. Use the highest NETStandard.Library available: This gets you the maximum number of fixes in NS.Library itself

cc @davidfowl

Issue #301, opened 2017-04-29
Address package downgrade warnings


mikeharder:
The latest build of `UniverseCoherence\2.0.0-preview1` (24826) contains package downgrade warnings in this repo.  All downgrade warnings are related to `StackExchange.Redis.StrongName`.

http://aspnetci/viewLog.html?buildId=214456&tab=buildResultsDiv&buildTypeId=UniverseCoherence

```
Detected package downgrade: System.Collections from 4.3.0 to 4.0.11  [C:\b\w\590aeb6bc9b87cb8\.r\Caching\.build\KoreBuild.proj]
 RedisCacheSample (>= 1.0.0) -> Microsoft.Extensions.Caching.Redis (>= 2.0.0-preview1-24822) -> StackExchange.Redis.StrongName (>= 1.1.605) -> System.IO.FileSystem (>= 4.0.1) -> runtime.win.System.IO.FileSystem (>= 4.3.0) -> System.Collections (>= 4.3.0)  [C:\b\w\590aeb6bc9b87cb8\.r\Caching\.build\KoreBuild.proj]
 RedisCacheSample (>= 1.0.0) -> Microsoft.Extensions.Caching.Redis (>= 2.0.0-preview1-24822) -> StackExchange.Redis.StrongName (>= 1.1.605) -> System.Collections (>= 4.0.11) [C:\b\w\590aeb6bc9b87cb8\.r\Caching\.build\KoreBuild.proj]
Detected package downgrade: System.IO from 4.3.0 to 4.1.0  [C:\b\w\590aeb6bc9b87cb8\.r\Caching\.build\KoreBuild.proj]
 RedisCacheSample (>= 1.0.0) -> Microsoft.Extensions.Caching.Redis (>= 2.0.0-preview1-24822) -> StackExchange.Redis.StrongName (>= 1.1.605) -> System.IO.FileSystem (>= 4.0.1) -> runtime.win.System.IO.FileSystem (>= 4.3.0) -> System.IO (>= 4.3.0)  [C:\b\w\590aeb6bc9b87cb8\.r\Caching\.build\KoreBuild.proj]
[...]

PR #304, opened 2017-05-08
Upgrade StackExchange.Redis.StrongName to version 1.2.3

mikeharder:
- Addresses package downgrade warnings (https://github.com/aspnet/Caching/issues/301)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ branches:
     - release
     - dev
     - /^(.*\/)?ci-.*$/
+    - /^rel\/.*/
 before_install:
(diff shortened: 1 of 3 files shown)


mikeharder:
@kichalla: Moved PR to preview1.  I will create a PR for BuildTools as well.

.travis.yml
```diff
@@ -16,6 +16,7 @@ branches:
     - release
     - dev
     - /^(.*\/)?ci-.*$/
+    - /^rel\/.*/
```
kichalla: 👍 We should add a work item to track making these changes in all our repos as this would be the usual format for future releases too

Issue #311, opened 2017-05-27
建议增加按前缀清理缓存的接口


molinjinyi:
```
public static void KeyDeleteWithPrefix(this IDatabase database, string prefix)
        {
            if (database == null)
            {
                throw new ArgumentException("Database cannot be null", "database");
            }

            if (string.IsNullOrWhiteSpace(prefix))
            {
                throw new ArgumentException("Prefix cannot be empty", "database");
            }

            database.ScriptEvaluate(@"
                local keys = redis.call('keys', ARGV[1]) 
                for i=1,#keys,5000 do 
                redis.call('del', unpack(keys, i, math.min(i+4999, #keys)))
                end", values: new RedisValue[] {prefix});
        }
```

```
public void Clear()
        {
            Database.KeyDeleteWithPrefix(GetLocalizedKey("*"));
        }
```

对于列表数据，你很难用明确的key来定位删除，没有这个接口缓存是不完美的

```
// Copyright (c) .NET Foundation. All rights reserved.
// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.

using System.Threading.Tasks;

namespace Microsoft.Extensions.Caching.Distributed
{
    public interface IDistributedCache
    {
        byte[] Get(string key);

        Task<byte[]> GetAsync(string key);

        void Set(string key, byte[] value, DistributedCacheEntryOptions options);

        Task SetAsync(string key, byte[] value, DistributedCacheEntryOptions options);

        void Refresh(string key);

        Task RefreshAsync(string key);

        void Remove(string key);

        Task RemoveAsync(string key);
    }
}
```
或者把连接暴露出来，让开发者更自由的使用

sandorfr:
See it seems to be the official issue to talk about the use cases around prefix based operation. It would be a good idea to translate the issue in english :) (yes I know about google translate ^^).

As I have a similar need, I'd like to share where I stand today. 

# Use case
I have a tree which takes a lot of computing to be built and which is different for each user depending on there permissions (ACL on that tree). Currently I build a full tree whenever it's invalidated and put it in cache. When a user requests it, I'd filter it based on his permissions and put it in cache so further requests would be fast.

When the tree is update, I would need to expire all user specific keys. 

## First Idea

[...]

PR #312, opened 2017-05-27
add clear count method extensions

molinjinyi:
add clear count method extensions


Tratcher:

PR #313, opened 2017-05-27
add clear cache method by prefix

molinjinyi:


Tratcher:

PR #314, opened 2017-05-27
add Clear cache

molinjinyi:
public void Clear(string prefix)
        {
            _cache.KeyDeleteWithPrefix(prefix+"*");
        }


Tratcher:

PR #318, opened 2017-06-14
Allow selecting database number and use existing IConnectionMultiplexer 

israellot:
Redis have a database concept for separating key spaces. It basically let us use usually up to 16 different databases on same instance. This allows better flexibility and better separation of key spaces instead of prefixing keys. 

I also introduced the possibility to inject an existing IConnectionMultiplexer for scenarios where the user has a singleton instance to be shared and minimize connections.

### DIFF
--- a/build/common.props
+++ b/build/common.props
@@ -16,8 +16,8 @@
     <PackageReference Include="Internal.AspNetCore.Sdk" Version="$(InternalAspNetCoreSdkVersion)" PrivateAssets="All" />
   </ItemGroup>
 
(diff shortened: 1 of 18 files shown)


muratg:
Thanks @israellot for submitting the PR.  Unfortunately, the change in `RedisCacheOptions` will be a breaking change that we can take in a minor release, so I'll close this one.  We'll use your bug https://github.com/aspnet/Caching/issues/343 to track this work.

israellot:
Thank you for considering it. Glad to help

Issue #325, opened 2017-06-29
Limit number of items in memory cache


JunTaoLuo:
We need to add an optional count limit on the memory cache to ensure a there is a maximum limit on the memory use of `MemoryCache`. When the count limit is reached, we will trim a percentage of the items in the cache (we have used 10% in the past). 

cc @muratg @Eilon 

Post 2.0 we plan to include a size hint to the memory cache and design more extensible eviction policies.

JunTaoLuo:
Followup item filed in #326

JunTaoLuo:
We decided to implement #326 instead.

PR #327, opened 2017-06-29
Add option to MemoryCacheOptions to specify maximum cache entries count

JunTaoLuo:
Addresses #325


src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -148,6 +150,12 @@ private void SetEntry(CacheEntry entry)
                 if (entryAdded)
                 {
                     entry.AttachTokens();
+
+                    // Compact by 10 percent if we exceed the given maximum number of cache entries
+                    if (_entries.Count > _maximumEntriesCount)
+                    {
+                        Compact(0.10);
```
JunTaoLuo: Do this in the background and handle concurrent adds.

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -148,6 +150,12 @@ private void SetEntry(CacheEntry entry)
                 if (entryAdded)
                 {
                     entry.AttachTokens();
+
+                    // Compact by 10 percent if we exceed the given maximum number of cache entries
+                    if (_entries.Count > _maximumEntriesCount)
```
Tratcher: I assume this always resolves to false for a nullable?

src/Microsoft.Extensions.Caching.Memory/MemoryCacheOptions.cs
```diff
@@ -13,6 +13,8 @@ public class MemoryCacheOptions : IOptions<MemoryCacheOptions>
 
         public TimeSpan ExpirationScanFrequency { get; set; } = TimeSpan.FromMinutes(1);
 
+        public int? EntryCountLimit { get; set; }
```
Tratcher: doc comment

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
JunTaoLuo: when it's null yes.

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -280,14 +281,36 @@ private static void ScanForExpiredItems(MemoryCache cache)
...
+                // Another compaction is running, exit immediately.
+                // Avoid overpurging when multiple overcapacity compactions are triggered concurrently.
+                return;
+            }
+
+            try
+            {
+                // Keep compacting until remaining entries are at 90% of maximum
+                // Stop compacting if no entries were removed, for example if all the remaining entries are pinned with NeverRemove priority
+                while (Compact(1 - ((0.9 * _entryCountLimit.Value) / _entries.Count)) && _entries.Count > _entryCountLimit.Value) { }
```
Tratcher: This was triggered by already reaching the max, so why not just ask to compact 10%?

No need for the loop, if it fills up again it will trigger compaction again.

JunTaoLuo: Stress testing it was failing. For example the test was using two threads to keep adding entries and the removal was not able to keep up. By the time 10% were trimmed, more than 10% were added. So instead, trim until 90% is left and keep doing this until 90% of max is achieved.

test/Microsoft.Extensions.Caching.Memory.Tests/CompactTests.cs
```diff
@@ -104,6 +105,9 @@ public void CompactWhenMaximumEntriesCountExceeded()
 
             cache.Set("key9", "value9");
 
+            // Wait 1 second for compaction to complete
+            Thread.Sleep(TimeSpan.FromSeconds(1));
```
Tratcher: register an eviction callback?

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
Tratcher: yikes, this loop may never exit and your cache could still keep growing beyond the limit. 
plan b: if we're at the limit when you try to add, discard the new item and invoke compaction. This should prevent you from almost ever going over the limit, let alone running to catch up. Make sure to run the normal callbacks for the new item.

JunTaoLuo: In that case, we are no longer trying to remove the least recently used entries. But I suppose that is still better than exceeding the memory limit.

Tratcher: The compaction will try to remove the least recently used to make room for the next guy, but in the meantime add must fail or block, and silent fail is better for a cache.

JunTaoLuo: Trying this now. This gives the extra benefit that the signature of Compact doesn't need to change.

JunTaoLuo: Seems to work but the behaviour is very strange. When the capacity is hit, the newest item and the oldest items(s) are evicted, whereas I would expect the newest one should be added and only the oldest items are removed.

JunTaoLuo:
Closed in favour of #332

PR #331, opened 2017-07-04
Remove redundant api check files

JunTaoLuo:
Noticed this when I was working in this repo. Since this is unrelated to the actual PR, I wanted to check this in first to reduce the spam for reviewers.


JunTaoLuo:
Oops approval I saw wasn't for this. But this change is pretty safe, let me know if this should be reverted.

Issue #326, opened 2017-06-29
Size hint based eviction for MemoryCache


JunTaoLuo:
Following up on https://github.com/aspnet/Caching/issues/325. 

`MemoryCache` does not know the size of the items it caches but the user of `MemoryCache` may know the size. In those cases we want to provide a method for size based eviction to guarantee a limit on the memory footprint of the cache. We can also add an option to require size hint when adding an item to the memory cache. Having a size hint also gives us the opportunity to investigate different eviction policies.

Tratcher:
Recommended for 2.1

javiercn:
Some notes from our discussion:
* Make sure we don't register MemoryDistributeCache on our components.
* Deprecate/remove the AddDistributedMemoryCache extension method. (Also update the comment to reflect that if this method is used, the cache will be shared among all IDistributedCache consumers).
* Force a size limit on DistributedMemoryCache
* Have each of our components on the framework that consume IDistributedCache create their own private version of the DistributedMemoryCache if no implementation is registered. The configuration on the limits for these instances can be exposed through named options or through some wrapper/new extension method on a case by case basis.
* Have the MemoryDistributedCache tell the size of the entries to the underlying MemoryCache when adding entries.

Sample `new extension method` for session:
```C#
services.AddInMemorySession(options => options.MaxSize = 10 * 1024 * 1024);
```

Sample `configuration through named options` for the DistributedCacheTagHelper
```C#
services.AddMvc();
services.Configure(Some.Constant, options => options.MaxSize = 10 * 1024 * 1024);
```

@Tratcher @davidfowl @JunTaoLuo

JunTaoLuo:
@javiercn the updated POR is collected in #336

Issue #329, opened 2017-07-03
Add MemoryDistributedCacheOptions to enables configuration of the MemoryCache used


javiercn:
https://github.com/aspnet/Caching/blob/dev/src/Microsoft.Extensions.Caching.Memory/MemoryCacheServiceCollectionExtensions.cs#L83

It makes it impossible to configure it.

muratg:
cc @davidfowl

JunTaoLuo:
POR:

- The MemoryDistributedCacheOptions should mirror the options available on MemoryCacheOptions
- A default size of 200 MB will be set
- The MemoryDistributedCache will by default create a new MemoryCache using the configured MemoryDistributedCacheOptions.

PR #332, opened 2017-07-04
Add size based eviction to MemoryCache

JunTaoLuo:
Addresses #326 

TODO:
- [x] Doc comments
- [x] Naming
- [x] Add more tests
- [ ] Evaluate perf impact
- [x] Investigate appveyor and travis failures

### DIFF
--- a/Caching.sln
+++ b/Caching.sln
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.26228.9
+VisualStudioVersion = 15.0.26621.2
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Extensions.Caching.Memory", "src\Microsoft.Extensions.Caching.Memory\Microsoft.Extensions.Caching.Memory.csproj", "{966D16D8-5D4E-4433-9DA7-F53EE44B7EE7}"
 EndProject
@@ -242,5 +242,8 @@ Global
 		{CD2A826A-A474-4491-94C8-43C557C68067} = {459E1593-2C11-42CB-AD17-F7597E69E5D2}
 		{17E332EB-D18D-4BF5-BCA5-989E36C78B79} = {459E1593-2C11-42CB-AD17-F7597E69E5D2}
 		{ADF83AC7-3776-4E62-A222-C6979C32DD2D} = {9E78AA8E-7870-46DE-A49F-856F5A0A9166}
+	EndGlobalSection
+	GlobalSection(ExtensibilityGlobals) = postSolution
+		SolutionGuid = {30FC4CB3-9789-4A49-939D-03398DBE03E0}
 	EndGlobalSection
 EndGlobal
--- a/src/Microsoft.Extensions.Caching.Abstractions/CacheEntryExtensions.cs
+++ b/src/Microsoft.Extensions.Caching.Abstractions/CacheEntryExtensions.cs
@@ -134,6 +134,24 @@ namespace Microsoft.Extensions.Caching.Memory
         }
 
         /// <summary>
+        /// Sets the size of the cache entry value.
+        /// </summary>
+        /// <param name="entry"></param>
+        /// <param name="size"></param>
+        public static ICacheEntry SetSize(
+            this ICacheEntry entry,
+            long size)
+        {
+            if (size < 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(size), size, $"{nameof(size)} must be non-negative.");
+            }
+
+            entry.Size = size;
+            return entry;
+        }
+
+        /// <summary>
         /// Applies the values of an existing <see cref="MemoryCacheEntryOptions"/> to the entry.
         /// </summary>
         /// <param name="entry"></param>
@@ -149,6 +167,7 @@ namespace Microsoft.Extensions.Caching.Memory
             entry.AbsoluteExpirationRelativeToNow = options.AbsoluteExpirationRelativeToNow;
             entry.SlidingExpiration = options.SlidingExpiration;
             entry.Priority = options.Priority;
+            entry.Size = options.Size;
 
             foreach (var expirationToken in options.ExpirationTokens)
             {
--- a/src/Microsoft.Extensions.Caching.Abstractions/MemoryCacheEntryExtensions.cs
+++ b/src/Microsoft.Extensions.Caching.Abstractions/MemoryCacheEntryExtensions.cs
@@ -22,6 +22,24 @@ namespace Microsoft.Extensions.Caching.Memory
         }
 
         /// <summary>
+        /// Sets the size of the cache entry value.
+        /// </summary>
+        /// <param name="options"></param>
+        /// <param name="size"></param>
+        public static MemoryCacheEntryOptions SetSize(
+            this MemoryCacheEntryOptions options,
+            long size)
+        {
+            if (size < 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(size), size, $"{nameof(size)} must be non-negative.");
+            }
+
+            options.Size = size;
+            return options;
+        }
+
+        /// <summary>
         /// Expire the cache entry if the given <see cref="IChangeToken"/> expires.
         /// </summary>
         /// <param name="options">The <see cref="MemoryCacheEntryOptions"/>.</param>
--- a/src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
+++ b/src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
@@ -19,6 +19,7 @@ namespace Microsoft.Extensions.Caching.Memory
     public class MemoryCache : IMemoryCache
     {
         private readonly ConcurrentDictionary<object, CacheEntry> _entries;
+        private long _cacheSize = 0;
         private bool _disposed;
 
         // We store the delegates locally to prevent allocations
@@ -26,9 +27,7 @@ namespace Microsoft.Extensions.Caching.Memory
         private readonly Action<CacheEntry> _setEntry;
         private readonly Action<CacheEntry> _entryExpirationNotification;
 
-        private readonly ISystemClock _clock;
-
-        private TimeSpan _expirationScanFrequency;
+        private readonly MemoryCacheOptions _options;
         private DateTimeOffset _lastExpirationScan;
 
         /// <summary>
@@ -42,14 +41,18 @@ namespace Microsoft.Extensions.Caching.Memory
                 throw new ArgumentNullException(nameof(optionsAccessor));
             }
 
-            var options = optionsAccessor.Value;
+            _options = optionsAccessor.Value;
+
             _entries = new ConcurrentDictionary<object, CacheEntry>();
             _setEntry = SetEntry;
             _entryExpirationNotification = EntryExpired;
 
-            _clock = options.Clock ?? new SystemClock();
-            _expirationScanFrequency = options.ExpirationScanFrequency;
-            _lastExpirationScan = _clock.UtcNow;
+            if (_options.Clock == null)
+            {
+                _options.Clock = new SystemClock();
+            }
+
+            _lastExpirationScan = _options.Clock.UtcNow;
         }
 
         /// <summary>
@@ -67,6 +70,9 @@ namespace Microsoft.Extensions.Caching.Memory
         {
             get { return _entries.Count; }
         }
+
+        // internal for testing
+        internal long Size { get => Interlocked.Read(ref _cacheSize); }
 
         private ICollection<KeyValuePair<object, CacheEntry>> EntriesCollection => _entries;
 
@@ -90,7 +96,12 @@ namespace Microsoft.Extensions.Caching.Memory
                 return;
             }
 
-            var utcNow = _clock.UtcNow;
+            if (_options.SizeLimit.HasValue && !entry.Size.HasValue)
+            {
+                throw new InvalidOperationException($"Cache entry must specify a value for {nameof(entry.Size)} when {nameof(_options.SizeLimit)} is set.");
+            }
+
+            var utcNow = _options.Clock.UtcNow;
 
             DateTimeOffset? absoluteExpiration = null;
             if (entry._absoluteExpirationRelativeToNow.HasValue)
@@ -122,7 +133,9 @@ namespace Microsoft.Extensions.Caching.Memory
                 priorEntry.SetExpired(EvictionReason.Replaced);
             }
 
-            if (!entry.CheckExpired(utcNow))
+            var exceedsCapacity = UpdateCacheSizeExceedsCapacity(entry);
+
+            if (!entry.CheckExpired(utcNow) && !exceedsCapacity)
             {
                 var entryAdded = false;
 
@@ -136,7 +149,15 @@ namespace Microsoft.Extensions.Caching.Memory
                     // Try to update with the new entry if a previous entries exist.
                     entryAdded = _entries.TryUpdate(entry.Key, entry, priorEntry);
 
-                    if (!entryAdded)
+                    if (entryAdded)
+                    {
+                        if (_options.SizeLimit.HasValue)
+                        {
+                            // The prior entry was removed, decrease the by the prior entry's size
+                            Interlocked.Add(ref _cacheSize, -priorEntry.Size.Value);
+                        }
+                    }
+                    else
                     {
                         // The update will fail if the previous entry was removed after retrival.
                         // Adding the new entry will succeed only if no entry has been added since.
@@ -151,6 +172,11 @@ namespace Microsoft.Extensions.Caching.Memory
                 }
                 else
                 {
+                    if (_options.SizeLimit.HasValue)
+                    {
+                        // Entry could not be added, reset cache size
+                        Interlocked.Add(ref _cacheSize, -entry.Size.Value);
+                    }
                     entry.SetExpired(EvictionReason.Replaced);
                     entry.InvokeEvictionCallbacks();
                 }
@@ -162,6 +188,14 @@ namespace Microsoft.Extensions.Caching.Memory
             }
             else
             {
+                if (exceedsCapacity)
+                {
+                    // The entry was not added due to overcapacity
+                    entry.SetExpired(EvictionReason.Capacity);
+
+                    TriggerOvercapacityCompaction();
+                }
+
                 entry.InvokeEvictionCallbacks();
                 if (priorEntry != null)
                 {
@@ -183,7 +217,7 @@ namespace Microsoft.Extensions.Caching.Memory
             CheckDisposed();
 
             result = null;
-            var utcNow = _clock.UtcNow;
+            var utcNow = _options.Clock.UtcNow;
             var found = false;
 
             CacheEntry entry;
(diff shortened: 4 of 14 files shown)


src/Microsoft.Extensions.Caching.Memory/MemoryCacheOptions.cs
```diff
@@ -13,6 +13,12 @@ public class MemoryCacheOptions : IOptions<MemoryCacheOptions>
 
         public TimeSpan ExpirationScanFrequency { get; set; } = TimeSpan.FromMinutes(1);
 
+        public bool EnforceSizeLimit { get; set; } = false;
+
+        public long HighWatermark { get; set; }
```
JunTaoLuo: Potentially extracted these options to a separate class/policy.

Tratcher: Naming, docs.
Make it nullable rather than having a separate EnforceSizeLimit bool?

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -42,14 +42,36 @@ public MemoryCache(IOptions<MemoryCacheOptions> optionsAccessor)
...
             }
 
-            var options = optionsAccessor.Value;
+            _options = optionsAccessor.Value;
+
+            if (_options.EnforceSizeLimit)
+            {
+                if (_options.HighWatermark <= 0)
+                {
+                    throw new ArgumentOutOfRangeException(nameof(_options.HighWatermark), _options.HighWatermark, $"{nameof(_options.HighWatermark)} must be positive.");
```
Tratcher: Move the validation to the options class.

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -90,7 +115,19 @@ private void SetEntry(CacheEntry entry)
...
-            var utcNow = _clock.UtcNow;
+            if (_options.EnforceSizeLimit)
+            {
+                if (!entry.Size.HasValue)
+                {
+                    throw new InvalidOperationException($"Cache entry must specify a value for {nameof(entry.Size)} when {nameof(_options.EnforceSizeLimit)} is set to true.");
+                }
+                if (entry.Size.Value < 0)
+                {
+                    throw new ArgumentOutOfRangeException(nameof(entry.Size.Value), entry.Size.Value, $"{nameof(entry.Size.Value)} must be non-negative.");
```
Tratcher: Move this validation to the property

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -90,7 +115,19 @@ private void SetEntry(CacheEntry entry)
                 return;
             }
 
-            var utcNow = _clock.UtcNow;
+            if (_options.EnforceSizeLimit)
```
Tratcher: Prevent adding the entry if doing so would exceed the cache size limit.

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -272,6 +328,33 @@ private static void ScanForExpiredItems(MemoryCache cache)
...
+                // Avoid overpurging when multiple overcapacity compactions are triggered concurrently.
+                return;
+            }
+
+            ThreadPool.QueueUserWorkItem(new WaitCallback(_ =>
+            {
+                try
+                {
+                    var currentSize = Interlocked.Read(ref _cacheSize);
+                    while (currentSize > _options.HighWatermark)
```
Tratcher: No loop. Try again next time.

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -281,18 +364,31 @@ private static void ScanForExpiredItems(MemoryCache cache)
...
+            int removalCountTarget = (int)(_entries.Count * percentage);
+            Compact(removalCountTarget, _ => 1);
+        }
+
+        private void Compact(long removalSizeTarget)
+        {
+            Compact(removalSizeTarget, entry => entry.Size.Value);
+        }
+
+        private void Compact(long removalSizeTarget, Func<CacheEntry, long> computeEntrySize)
```
Tratcher: Let's discuss

src/Microsoft.Extensions.Caching.Memory/MemoryCacheServiceCollectionExtensions.cs
```diff
@@ -80,9 +80,44 @@ public static IServiceCollection AddDistributedMemoryCache(this IServiceCollecti
...
+
+            if (setupAction == null)
+            {
+                throw new ArgumentNullException(nameof(services));
+            }
+
+            var memoryCacheOptions = new MemoryCacheOptions();
+            setupAction(memoryCacheOptions);
+
+            services.TryAddSingleton<IDistributedCache>(new MemoryDistributedCache(new MemoryCache(memoryCacheOptions)));
```
Tratcher: https://github.com/aspnet/Caching/issues/329 ?
I don't think this was the fix they had in mind.

Tratcher:
MemoryDistributedCache should set sizes by default. This will enable Session to work without changes.

Caching.sln
```diff
@@ -243,4 +243,7 @@ Global
 		{17E332EB-D18D-4BF5-BCA5-989E36C78B79} = {459E1593-2C11-42CB-AD17-F7597E69E5D2}
 		{ADF83AC7-3776-4E62-A222-C6979C32DD2D} = {9E78AA8E-7870-46DE-A49F-856F5A0A9166}
 	EndGlobalSection
+	GlobalSection(ExtensibilityGlobals) = postSolution
```
JunTaoLuo: Strange, not sure why vs added this.

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -142,6 +158,14 @@ private void SetEntry(CacheEntry entry)
                         // Adding the new entry will succeed only if no entry has been added since.
                         // This guarantees removing an old entry does not prevent adding a new entry.
                         entryAdded = _entries.TryAdd(entry.Key, entry);
+
+                        if (entryAdded)
+                        {
+                            // We have only added the difference between the new size and the old size.
+                            // Since the old one was removed, we need to add the old entry's size back
+                            // to ensure the total size of the new entry is accounted for.
+                            Interlocked.Add(ref _cacheSize, priorEntry.Size.Value);
```
JunTaoLuo: This could lead to a situation where the capacity is exceeded once SetEntry is complete. Will look for another way to accomplish this.

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -122,7 +136,9 @@ private void SetEntry(CacheEntry entry)
                 priorEntry.SetExpired(EvictionReason.Replaced);
```
Tratcher: Why not deduct the size immediately? Then you can deal with just the new size.

JunTaoLuo: I keep the size operations beside the actual calls to remove from the underlying concurrent dictionary. In this case, that's the ScanForExpiredItems call. I think we decided not to remove the entry right away to ensure the value can never be null on replacement hence the awkward logic here. I don't think it's safe to deduct the size here.

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -162,6 +191,25 @@ private void SetEntry(CacheEntry entry)
...
+                if (_options.SizeLimit != null && updatedCacheSize > _options.SizeLimit.High)
+                {
+                    // The entry was not added due to overcapacity
+                    entry.SetExpired(EvictionReason.Capacity);
+
+                    // The cache size needs to be reset
+                    Interlocked.Add(ref _cacheSize, -sizeUpdate);
+
+                    // Spawn background thread for compaction
+                    ThreadPool.QueueUserWorkItem(new WaitCallback(_ =>
```
JunTaoLuo: I used to have one but this was the only place that was calling it.

davidfowl: Can you remove the closure? Seems like you're not capturing any local state.

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -272,6 +321,34 @@ private static void ScanForExpiredItems(MemoryCache cache)
...
+            sizeUpdate = priorEntry == null ? entry.Size.Value : entry.Size.Value - priorEntry.Size.Value;
+            updatedCacheSize = Interlocked.Add(ref _cacheSize, sizeUpdate);
+
+            return updatedCacheSize > _options.SizeLimit;
+        }
+
+        private void TriggerOvercapacityCompaction()
+        {
+            // Spawn background thread for compaction
+            ThreadPool.QueueUserWorkItem(new WaitCallback(_ =>
```
davidfowl: Remove the new WaitCallback.

src/Microsoft.Extensions.Caching.Memory/MemoryCacheOptions.cs
```diff
@@ -9,10 +9,41 @@ namespace Microsoft.Extensions.Caching.Memory
...
     public class MemoryCacheOptions : IOptions<MemoryCacheOptions>
     {
+        private long? _sizeLimit;
+        private double _removalPercentageOnOvercapacityCompaction = 0.05;
+
         public ISystemClock Clock { get; set; }
 
         public TimeSpan ExpirationScanFrequency { get; set; } = TimeSpan.FromMinutes(1);
 
+        public long? SizeLimit
```
davidfowl: I don't want to do a back and forth but forcing users to calculate a percentage seems bizarre.

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -90,7 +96,15 @@ private void SetEntry(CacheEntry entry)
                 return;
             }
 
-            var utcNow = _clock.UtcNow;
+            if (_options.SizeLimit != null)
```
Tratcher: .HasValue everywhere

src/Microsoft.Extensions.Caching.Abstractions/MemoryCacheEntryExtensions.cs
```diff
@@ -21,6 +21,20 @@ public static class MemoryCacheEntryExtensions
             return options;
         }
 
+        // TODO: Doc comments
+        public static MemoryCacheEntryOptions SetSize(
+            this MemoryCacheEntryOptions options,
+            long size)
+        {
+            if (size < 0)
```
muratg: Does size==0 signify anything?  Also, is this "Size in bytes"?

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -272,6 +321,34 @@ private static void ScanForExpiredItems(MemoryCache cache)
...
+            return updatedCacheSize > _options.SizeLimit;
+        }
+
+        private void TriggerOvercapacityCompaction()
+        {
+            // Spawn background thread for compaction
+            ThreadPool.QueueUserWorkItem(new WaitCallback(_ =>
+            {
+                var currentSize = Interlocked.Read(ref _cacheSize);
+                if (currentSize > _options.SizeLimit * (1 - _options.RemovalPercentageOnOvercapacityCompaction))
```
Tratcher: save your calculation in a local variable to avoid repetition.

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -281,18 +358,31 @@ private static void ScanForExpiredItems(MemoryCache cache)
         /// ?. Larger objects - estimated by object graph size, inaccurate.
         public void Compact(double percentage)
         {
+            int removalCountTarget = (int)(_entries.Count * percentage);
+            Compact(removalCountTarget, _ => 1);
+        }
+
+        private void Compact(long removalSizeTarget)
```
Tratcher: this helper method isn't doing anything, do you need it?

src/Microsoft.Extensions.Caching.Abstractions/MemoryCacheEntryExtensions.cs
JunTaoLuo: We don't want to impose what size means. When we use this in response caching, it will be bytes but in theory you can have a memory cache where all the sizes are 1 and in that case size means count.

JunTaoLuo: Size of zero means you can add this entry even when the cache is full.

src/Microsoft.Extensions.Caching.Memory/MemoryCacheOptions.cs
JunTaoLuo: Let's talk about this in person

Tratcher: Calculate a percentage? What are their inputs? "If my cache is over capacity I need to purge some data. How much data do I purge?" I'll argue that they'll be more likely to start with a percentage than a fixed value, "free up 5% capacity", which is much simpler to reason about than to say "how big is my total cache? 1gb? Free up 50mb".

src/Microsoft.Extensions.Caching.Abstractions/CacheEntryExtensions.cs
```diff
@@ -133,6 +133,20 @@ public static class CacheEntryExtensions
             return entry;
         }
 
+        // TODO: Add doc comments
+        public static ICacheEntry SetSize(
+            this ICacheEntry entry,
+            long size)
+        {
+            if (size < 0)
```
JunTaoLuo: Add some tests for these checks

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -162,6 +188,17 @@ private void SetEntry(CacheEntry entry)
             }
             else
             {
+                if (_options.SizeLimit.HasValue && updatedCacheSize > _options.SizeLimit)
```
Tratcher: isn't this just `exceedsCapacity`?

src/Microsoft.Extensions.Caching.Memory/MemoryCacheOptions.cs
```diff
@@ -9,10 +9,41 @@ namespace Microsoft.Extensions.Caching.Memory
...
+                if (value <= 0)
+                {
+                    throw new ArgumentOutOfRangeException(nameof(value), value, $"{nameof(value)} must be positive.");
+                }
+
+                _sizeLimit = value;
+            }
+        }
+
+        public double RemovalPercentageOnOvercapacityCompaction
```
Tratcher: `CompactionPercentage`?

src/Microsoft.Extensions.Caching.Memory/MemoryCacheOptions.cs
```diff
@@ -9,10 +9,41 @@ namespace Microsoft.Extensions.Caching.Memory
...
         public ISystemClock Clock { get; set; }
 
         public TimeSpan ExpirationScanFrequency { get; set; } = TimeSpan.FromMinutes(1);
 
+        public long? SizeLimit
+        {
+            get => _sizeLimit;
+            set
+            {
+                if (value <= 0)
```
Tratcher: allow 0 for test scenarios. e.g. temporarily reduce the cache limit to 0 to debug an issue masked (or caused) by caching.

src/Microsoft.Extensions.Caching.Abstractions/CacheEntryExtensions.cs
```diff
@@ -134,6 +134,24 @@ public static class CacheEntryExtensions
         }
 
         /// <summary>
+        /// Sets the size of the cache entry.
```
Tratcher: the cache entry Value.
The thing being cached, not the infrastructure around it.

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -272,6 +321,53 @@ private static void ScanForExpiredItems(MemoryCache cache)
             }
         }
 
+        private bool UpdateCacheSizeExceedsCapacity(CacheEntry entry, out bool cacheSizeUpdated)
```
Tratcher: Can you eliminate cacheSizeUpdated? e.g. if it exceed capacity, immediately subtract it again. Then the only info you need out of this method is the return bool.

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -272,6 +321,53 @@ private static void ScanForExpiredItems(MemoryCache cache)
...
+        {
+            if (!_options.SizeLimit.HasValue)
+            {
+                cacheSizeUpdated = false;
+                return false;
+            }
+
+            var newSize = 0L;
+            while(true)
```
Tratcher: finite retries? e.g. if it fails 100 times then give up and return that you're over capacity. Under stress you may get a bunch of threads in this loop competing with eachother and there's no gurantee how long they'll be spinning. Some poor threads could be stuck here forever burning all the CPU on your machine (unlikely but possible). MemoryCache needs to be fast, and that includes failing fast if necessary.

JunTaoLuo: Fair

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
JunTaoLuo: Possible, I'll take a closer look in the morning.

test/Microsoft.Extensions.Caching.Memory.Tests/CapacityTests.cs
```diff
@@ -0,0 +1,391 @@
...
+            }
+
+            // There should be 4 items in the cache
+            Assert.Equal(4, cache.Count);
+
+            cache.Set("key5", "value5", new MemoryCacheEntryOptions { Size = 1 });
+            testClock.Add(TimeSpan.FromSeconds(10));
+
+            // Wait for compaction to complete
+            Thread.Sleep(TimeSpan.FromSeconds(4));
```
JunTaoLuo: Increased these waits significantly to ensure tests pass on appveyor and travis. This is a rather poor approach and I'll see if there are better ways to ensure the consistency of these tests.

src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs
```diff
@@ -272,6 +315,57 @@ private static void ScanForExpiredItems(MemoryCache cache)
...
+                return false;
+            }
+
+            var newSize = 0L;
+            for (var i = 0; i < 100; i++)
+            {
+                var sizeRead = Interlocked.Read(ref _cacheSize);
+                newSize = sizeRead + entry.Size.Value;
+
+                if (newSize < 0)
```
Tratcher: Move `newSize > _options.SizeLimit` here and get rid of the decrement.

test/Microsoft.Extensions.Caching.Memory.Tests/CapacityTests.cs
```diff
@@ -0,0 +1,435 @@
...
+            Assert.Throws<ArgumentOutOfRangeException>(() => options.CompactionPercentage = 1.1);
+            Assert.Throws<ArgumentOutOfRangeException>(() => options.CompactionPercentage = -0.1);
+        }
+
+        [Fact]
+        public void AddingEntryIncreasesCacheSizeWhenEnforcingSizeLimit()
+        {
+            var cache = new MemoryCache(new MemoryCacheOptions
+            {
+                ExpirationScanFrequency = TimeSpan.Zero,
```
Tratcher: remove

Issue #320, opened 2017-06-15
Method not found - Microsoft.Extensions.Caching.Memory.MemoryCacheOptions.set_CompactOnMemoryPressure(Boolean)


darkengines:
# csproj

```xml
  <PropertyGroup>
    <TargetFramework>netcoreapp2.0</TargetFramework>
    <RuntimeIdentifier>win7-x86</RuntimeIdentifier>
    <RuntimeFrameworkVersion>2.0.0-preview2-25407-01</RuntimeFrameworkVersion>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="RazorLight" Version="1.1.0" />
    <PackageReference Include="Microsoft.Extensions.Caching.Abstractions" Version="2.0.0-preview3-25641" />
    <PackageReference Include="Microsoft.Extensions.Caching.Memory" Version="2.0.0-preview3-25641" />
  </ItemGroup>

```
I have https://dotnet.myget.org/F/dotnet-core/api/v3/index.json registered as a Nuget feed.

# Issue
This line:

`EngineFactory.CreatePhysical(Path.Combine(HostingEnvironment.ContentRootPath, @"Mailing\Templates"))`

(RazorLight template engine creation https://github.com/toddams/RazorLight)
Known issue : https://github.com/toddams/RazorLight/issues/69

Throws: 

> System.MissingMethodException : 'Method not found: 'Void Microsoft.Extensions.Caching.Memory.MemoryCacheOptions.set_CompactOnMemoryPressure(Boolean)'.'

Can you do something from your side?
No build error / warning.

Personal note: I think this is the last issue before we can move on netcoreapp2.0.

Tratcher:
CompactOnMemoryPressure was removed in 2.0. It sounds like RazorLight depends on 1.1.1. You won't be able to use it with the 2.0 packages until they update their dependency.

Tratcher:
@JunTaoLuo @muratg

JunTaoLuo:
We may want to bring back the options and mark it as obsolete so we don't break consumers when they switch to 2.0.0. We removed it assuming no one was depending which is obviously not the case. Otherwise, we can wait until all third parties react accordingly.

muratg:
@glennc Thoughts?  Do we want to bring back the options and add obsolete attributes?

muratg:
Small work item to ensure we don't break 3rd party libraries.  Assigning in 2.0.0 to @JunTaoLuo.

PR #333, opened 2017-07-07
Restore and obsolete the CompactOnMemoryPressure property

JunTaoLuo:
Addresses #320 

Do we need a fwlink for this?

### DIFF
--- a/src/Microsoft.Extensions.Caching.Memory/MemoryCacheOptions.cs
+++ b/src/Microsoft.Extensions.Caching.Memory/MemoryCacheOptions.cs
@@ -11,6 +11,9 @@ namespace Microsoft.Extensions.Caching.Memory
     {
         public ISystemClock Clock { get; set; }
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.Extensions.Caching.Memory/MemoryCacheOptions.cs
```diff
@@ -11,6 +11,9 @@ public class MemoryCacheOptions : IOptions<MemoryCacheOptions>
     {
         public ISystemClock Clock { get; set; }
 
+        [Obsolete("This is obsolete and will be removed in a future version.")]
+        public bool CompactOnMemoryPressure { get; set; } = true;
```
Tratcher: Want to change the default to false? :-) we don't do it anymore.

JunTaoLuo: Depends on how much of this we want to keep the same as before. Doesn't really matter either way.

PR #336, opened 2017-07-07
Add MemoryDistributedCacheOptions and use it to configure MemoryCache used by MemoryDistributedCache

JunTaoLuo:
Needs rebase once #332 goes in.

Addresse #329

### DIFF
--- a/src/Microsoft.Extensions.Caching.Memory/MemoryCacheServiceCollectionExtensions.cs
+++ b/src/Microsoft.Extensions.Caching.Memory/MemoryCacheServiceCollectionExtensions.cs
@@ -50,7 +50,7 @@ namespace Microsoft.Extensions.DependencyInjection
 
             if (setupAction == null)
             {
-                throw new ArgumentNullException(nameof(services));
+                throw new ArgumentNullException(nameof(setupAction));
             }
 
             services.AddMemoryCache();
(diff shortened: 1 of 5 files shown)


src/Microsoft.Extensions.Caching.Memory/MemoryCacheServiceCollectionExtensions.cs
```diff
@@ -80,7 +80,44 @@ public static IServiceCollection AddDistributedMemoryCache(this IServiceCollecti
...
+        /// server scenarios as this cache stores items in memory and doesn't expand across multiple machines.
+        /// For those scenarios it is recommended to use a proper distributed cache that can expand across
+        /// multiple machines.
+        /// </remarks>
+        /// <param name="services">The <see cref="IServiceCollection" /> to add services to.</param>
+        /// <param name="setupAction">
+        /// The <see cref="Action{MemoryDistributedCacheOptions}"/> to configure the provided <see cref="MemoryDistributedCacheOptions"/>.
+        /// </param>
+        /// <returns>The <see cref="IServiceCollection"/> so that additional calls can be chained.</returns>
+        public static IServiceCollection AddDistributedMemoryCache(this IServiceCollection services, Action<MemoryDistributedCacheOptions> setupAction)
```
JunTaoLuo: `AddDistributedMemoryCache` vs `MemoryDistributedCache` we should normalize the naming here. Pick one and go with it.

Tratcher: This is ok. Both names are already used publicly, you can't change it without breaking APIs.

JunTaoLuo: I think it's good to keep things consistent and if we are going to break things, we should break them in 2.0

Tratcher: *shrug* yesterday maybe.

JunTaoLuo: Well we are branching on Monday now so... it's functionally yesterday.

JunTaoLuo:
I'm going to merge as is for now, if we want to change the naming, we can do it afterwards.

JunTaoLuo:
merged in https://github.com/aspnet/Caching/commit/5e73a85ef5e6f1bf61f619c1c4c50437e898e616

PR #337, opened 2017-07-08
Implement IOptions accessor for MemoryDistributedCacheOptions

JunTaoLuo:
Something we should have done in #336


Tratcher:
You don't really need this. We only did it to MemoryCacheOptions so you could easily new up a MemoryCache (e.g. `new MemoryCache(new MemoryCacheOptions())`), because MemoryCache was something people were likely to create without DI. The same is not true for MemoryDistributedCache, it's only a DI fallback for a real IDistributedCache.

JunTaoLuo:
I noticed [this](https://github.com/aspnet/Session/blob/dev/test/Microsoft.AspNetCore.Session.Tests/SessionTests.cs#L724) in Session when I was fixing the build, but that was only a test. If we are sure there's no use case for newing up an instance of MemoryDistributedCache then this isn't a big deal.

JunTaoLuo:
Closing, if we need it, then we'll add it.

PR #354, opened 2017-10-03
Update doc comments

JunTaoLuo:
I noticed this as I was responding to an issue. We should match the doc comment to the actual property it is describing. Also, a time duration between successive events should be called a period and frequency = 1/period. Calling the property ExpirationScanFrequency is not technically accurate.


JunTaoLuo:
And also I opted to not write doc comments for the obsoleted option since we are going to remove it in the future anyway.

PR #364, opened 2017-12-01
Update bootstrappers

ryanbrandenburg:


ryanbrandenburg:
Failure appears to be an existing one. Merging.

Issue #316, opened 2017-06-13
CacheExtensions methods do not validate their arguments


anpete:
The public extension methods in MemoryCacheExtensions.cs (file name doesn't match class name) do not validate their arguments.

PR #369, opened 2018-02-24
Added suitable null checks to the extension method arguments

d-saravanan:
Summary of the changes (Less than 80 chars)
 - Added suitable null checks to the extension method arguments

Addresses #316

### DIFF
--- a/src/Microsoft.Extensions.Caching.Abstractions/MemoryCacheExtensions.cs
+++ b/src/Microsoft.Extensions.Caching.Abstractions/MemoryCacheExtensions.cs
@@ -12,18 +12,33 @@ namespace Microsoft.Extensions.Caching.Memory
 
         public static object Get(this IMemoryCache cache, object key)
         {
+            if (key == null)
+            {
+                throw new System.ArgumentNullException(nameof(key));
+            }
+
             cache.TryGetValue(key, out object value);
             return value;
         }
 
         public static TItem Get<TItem>(this IMemoryCache cache, object key)
         {
+            if (key == null)
+            {
+                throw new System.ArgumentNullException(nameof(key));
+            }
+
             cache.TryGetValue<TItem>(key, out TItem value);
             return value;
         }
 
         public static bool TryGetValue<TItem>(this IMemoryCache cache, object key, out TItem value)
         {
+            if (key == null)
+            {
+                throw new System.ArgumentNullException(nameof(key));
+            }
+
             if (cache.TryGetValue(key, out object result))
             {
                 value = (TItem)result;
@@ -36,6 +51,11 @@ namespace Microsoft.Extensions.Caching.Memory
 
         public static TItem Set<TItem>(this IMemoryCache cache, object key, TItem value)
         {
+            if (key == null)
+            {
+                throw new System.ArgumentNullException(nameof(key));
+            }
+
             var entry = cache.CreateEntry(key);
             entry.Value = value;
             entry.Dispose();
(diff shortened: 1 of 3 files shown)


d-saravanan:
Can anyone trigger a new build on this, because the build in my repository is passing and only here its failing

src/Microsoft.Extensions.Caching.Abstractions/MemoryCacheExtensions.cs
```diff
@@ -45,6 +65,10 @@ public static TItem Set<TItem>(this IMemoryCache cache, object key, TItem value)
 
         public static TItem Set<TItem>(this IMemoryCache cache, object key, TItem value, DateTimeOffset absoluteExpiration)
         {
+            if (key == null)
+            {
+                throw new System.ArgumentNullException(nameof(key));
```
Tratcher: Add a using for System.

src/Microsoft.Extensions.Caching.Abstractions/MemoryCacheExtensions.cs
```diff
@@ -12,18 +12,33 @@ public static class CacheExtensions
 
         public static object Get(this IMemoryCache cache, object key)
         {
+            if (key == null)
```
jkotalik: Can we add a few tests to see that a ArgumentNullException is thrown. See: https://github.com/aspnet/Caching/blob/dev/test/Microsoft.Extensions.Caching.Redis.Tests/RedisCacheSetAndRemoveTests.cs#L93-L97 for a good example.

d-saravanan: I have added 6 test cases to test all the different scenario's and have also refactored the MemoryCacheExtensions.cs [CacheExtensions] to remove the redundant validations made in the earlier check-in. Please review and share your review comments.

src/Microsoft.Extensions.Caching.Abstractions/MemoryCacheExtensions.cs
d-saravanan: I have updated the code and sent a new pull-request and raised a review.

src/Microsoft.Extensions.Caching.Abstractions/MemoryCacheExtensions.cs
```diff
@@ -119,5 +128,15 @@ public static TItem GetOrCreate<TItem>(this IMemoryCache cache, object key, Func
 
             return (TItem)result;
         }
+
+        #region Private Members
```
Tratcher: No regions please.

davidfowl:
I hope we don't double null check in the default scenario.

d-saravanan:
@davidfowl I have ensured there are no double null checks. However, if we can move the null check to MemoryCache.cs [CreateEntry method], we can remove the checks that are present in the MemoryCacheExtensions.cs [CacheExtensions class]. Please let me know if I should move the checks inside MemoryCache.

src/Microsoft.Extensions.Caching.Abstractions/MemoryCacheExtensions.cs
d-saravanan: @Tratcher : For my learning, can you please let me know why not to use regions. I have been using them in my source code. Is that not a good practice. Should I remove them and raise a new pull-request.

Tratcher: It's only a style issue, we don't use them in most cases, especially not for a single method.

jkotalik:
I think we can just add a null check here: https://github.com/aspnet/Caching/blob/dev/src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs#L83. This would simplify the check to one place and reduce redundancy.

d-saravanan:
@jkotalik I had earlier posted in the reply David Fowler. However, I will update and initiate a new pull-request.

d-saravanan:
@jkotalik I have updated and initiated a new pull-request. Kindly review the same and share your comments if any.

PR #378, opened 2018-03-30
Double check assertion fix

sbebrys:
Fix double check assertion because in multithreads _cache can be null
when _connection has been just created


Tratcher:
It's this theoretical or have you actually encountered an error here?

sbebrys:
Not from this place because first 'if' leave method and I have NullReferenceException when _cache is used in other place

JunTaoLuo:
Merged in https://github.com/aspnet/Caching/commit/02600b6e8b6a6101991a553447b833ad70175781. Thanks @sbebrys !

PR #394, opened 2018-05-23
Async extension methods for setting and getting values

dejanstojanovic:
SetAsync and GetAsync values that take generic type T rather than byte array for easier setting and retrieving values using distributed caching


Tratcher:
Thanks for the PR, but we do not plan on taking this change. We've intentionally kept serialization as an independent concern as there are several ways to go about it and they all have trade offs.

Issue #361, opened 2017-11-20
MemoryCache generic TryGetValue throws InvalidCastException when trying to get value


tstanitz:
Hello,

I have found that InvalidCastException coming from the generic TryGetValue extension method when the type of the value related to the key is not compatible with the provided generic type.

To reproduce:
1. Add a value to the cache for some key
2. Call the TryGetValue generic method with a key and a non-compatible out parameter
3. InvalidCastException is thrown

It should not throw any exception, but return false and set the out parameter to default.

The exception comes from the explicit casting:
https://github.com/aspnet/Caching/blob/23a89b78fedea611d74e3a754bc4843115013ae1/src/Microsoft.Extensions.Caching.Abstractions/MemoryCacheExtensions.cs

Thanks

Tratcher:
Do you really use different object types in the same cache key? It seems like that would get very confusing.

tstanitz:
I agree with you that it is and edge case, but I think it is not a good idea to throw exception from this method, therefore I have reported it.

JunTaoLuo:
I think we should still follow the convention of other APIs for Try* where if it's not successful, it should return false instead of throwing.

tstanitz:
I agree that when it doesn't throw InvalidCastException, but returns false then it is not able to differenciate the two cases when the key is not existing in the cache or it exists but not able to cast.

tstanitz:
As I see the ConcurrentDictionary TryGetValue(TKey key, out TValue value) and MemoryCache TryGetValue(object key, out object value) also throws exception (ArgumentNullException when key is null) but at least it is documented.

Could you please decide to change it to not throw exception just return false or update the document to mention the InvalidCastException or leave it as it is

Thanks in advance

PR #395, opened 2018-05-24
MemoryCache generic TryGetValue return false when incompatible type used

tstanitz:
Pull request for #361

### DIFF
--- a/src/Microsoft.Extensions.Caching.Abstractions/MemoryCacheExtensions.cs
+++ b/src/Microsoft.Extensions.Caching.Abstractions/MemoryCacheExtensions.cs
@@ -17,8 +17,7 @@ namespace Microsoft.Extensions.Caching.Memory
 
         public static TItem Get<TItem>(this IMemoryCache cache, object key)
         {
-            cache.TryGetValue(key, out TItem value);
-            return value;
+            return (TItem)cache.Get(key);
         }
 
         public static bool TryGetValue<TItem>(this IMemoryCache cache, object key, out TItem value)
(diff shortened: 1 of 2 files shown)


src/Microsoft.Extensions.Caching.Abstractions/MemoryCacheExtensions.cs
```diff
@@ -25,8 +25,11 @@ public static bool TryGetValue<TItem>(this IMemoryCache cache, object key, out T
         {
             if (cache.TryGetValue(key, out object result))
             {
-                value = (TItem)result;
-                return true;
+                if (result is TItem item)
```
Tratcher: This has some subtle side-effects. The behavior of `Get<TItem>` changes, but not `Get` or `GetOrCreate<TItem>` or `GetOrCreateAsync<TItem>`. Change `Get<TItem>` to throw?

tstanitz: oh, you are right. What do you suggest?

Tratcher: Change `Get<TItem>` to call `Get` and Cast instead of calling `TryGet<TItem>`.

Tratcher:
Those test failures are suspicious. It's null reffing on `cache.Get<int>(i);`. It can't cast null to an int. Before it would have returned default/0. `Get<TItem>` needs to change to something like `return (TItem)(cache.Get(key) ?? default);`

tstanitz:
I have realized in the meantime. I will update the pr. Thanks for your help

PR #398, opened 2018-05-30
Contains method added IMemoryCache, implemted MemoryCache and test methods added

yunuskiran:


src/Microsoft.Extensions.Caching.Abstractions/IMemoryCache.cs
```diff
@@ -30,5 +30,12 @@ public interface IMemoryCache : IDisposable
         /// </summary>
         /// <param name="key">An object identifying the entry.</param>
         void Remove(object key);
+
+        /// <summary>
+        /// Key exist in the cache
+        /// </summary>
+        /// <param name="key"></param>
+        /// <returns></returns>
+        bool Contains(object key);
```
Tratcher: Breaking change, we can't add methods to an interface in 2.2.

Why do you need this? It's functionally equivalent to TryGetValue.

yunuskiran: because i only ask the store is it contains. if is it contains, i create some special operations. TryGetValue get value.

Tratcher: The TryGetValue return value is the same as Contains. Getting the value back does not cost any more. Contains could be written as an extension method over TryGetValue.

yunuskiran: Thanks for information. I will apply this extension in my code.

Issue #429, opened 2018-10-12
Test failure: DoNotAddIfSizeOverflows


ryanbrandenburg:
This issue was made automatically. If there is a problem contact ryanbrandenburg.

This test [failed](https://dnceng.visualstudio.com/9ee6d478-d288-47f7-aacc-f6e6d082ae6d/_build/results?buildId=29613) with the following error:

```
Assert.True() Failure
Expected: True
Actual:   False
```

Other tests within that build may have failed with a similar message, but they are not listed here. Check the link above for more info.

This test failed on master.

CC @Eilon

ryanbrandenburg:
This comment was made automatically. If there is a problem contact ryanbrandenburg.

Please use this workflow to address this flaky test issue, including checking applicable checkboxes and filling in the applicable "TODO" entries:

* Is this actually a flaky test?
  * No, this is a regular test failure, fix the test/product (TODO: Link to commit/PR)
  * Yes, proceed below...

* Is this test failure caused by product code flakiness? (Either this product, or another product this test depends on.)
  * [ ] File a bug against the product (TODO: Link to other bug)
  * Is it possible to change the test to avoid the flakiness?
    * Yes? Go to the "Change the test!" section.
    * No?
      * [ ] Disable the test (TODO: Link to PR/commit)
      * [ ] Wait for other bug to be resolved
      * [ ] Wait for us to get build that has the fix
      * [ ] Re-enable our test (TODO: Link to PR/commit)
      * [ ] Close this bug

* Is it that the test itself is flaky? This includes external transient problems (e.g. remote server problems, file system race condition, etc.)
  * Is there is a way to change our test to avoid this flakiness?
    * Yes? Change the test!
      * [ ] Change the test to avoid the flakiness, for example by using a different test strategy, or by adding retries w/ timeouts (TODO: Link to PR/commit)
      * [ ] Run the test 100 times locally as a sanity check.
      * [ ] Close this bug
    * No?
      * Is there any logging or extra information that we could add to make this more diagnosable when it happens again?
        * Yes?
            * [ ] Add the logging (TODO: Link to PR/commit)
        * No?
            * [ ] Delete the test because flaky tests are not useful (TODO: Link to PR/commit)

Eilon:
@muratg - all yours.

muratg:
@JunTaoLuo can you take a look? Any changes which may have caused this?

aspnet-hello:
[...]

Issue #261, opened 2017-01-10
Add Logging


JunTaoLuo:
There's currently no default logging for cache operations. It would be useful to turn on logging and see logged cache operations.

JunTaoLuo:
@muratg suggest we punt this.

PR #441, opened 2018-11-15
Add logging for capacity tests

JunTaoLuo:
Adding logging to test out my theory on https://github.com/aspnet/Caching/issues/429. I opted for an internal property used for tests since it's simple. However, this wouldn't help with https://github.com/aspnet/Caching/issues/261.


JunTaoLuo:
Oops. I should open this on Extensions.

